Issue #1, opened 2021-10-21
Initial Lower DRB DO data inventory


jsadler2:
We need to know how much DO data there is available in the lower DRB

PR #10, opened 2021-11-04
add initial DO pipeline files

lekoenig:
This PR addresses issue #1. I've initialized the targets pipeline for this project, which currently downloads and filters DO data from the WQP and NWIS, and outputs some summary information and figures about data coverage. 

Any feedback or questions about the filtering steps or the DO data are welcome! Let's DO it!


1_fetch/src/find_nwis_sites.R
```diff
@@ -0,0 +1,19 @@
+get_drb_sites <- function(hucs,pcodes_select,site_tp_select){
+  
+  # hucs is a character vector containing the minor HUCs (hydrologic unit codes) over which to search. A minor HUC is 8 digits in length. 
+  # pcodes_select is a character vector containing the USGS parameter codes of interest
+  # site_tp_select is a character vector containing the USGS site types to retain
+  
+  # Search for sites separately by HUC8 region since calls to readNWISdata allow no more than 10 minor HUCs to be specified:
+  drb_sites_ls <- lapply(hucs,function(x)
+    readNWISdata(huc=x,parameterCd=pcodes_select,service="site",seriesCatalogOutput=TRUE))
```
jsadler2, line 9: I think you can include siteType in the service call which would let you filter upfront instead of with the lines below. https://waterservices.usgs.gov/rest/Site-Service.html#siteType

1_fetch/src/find_nwis_sites.R
```diff
@@ -0,0 +1,19 @@
+get_drb_sites <- function(hucs,pcodes_select,site_tp_select){
+  
+  # hucs is a character vector containing the minor HUCs (hydrologic unit codes) over which to search. A minor HUC is 8 digits in length.
```
jsadler2, line 3: Very helpful to have these explanations for the function parameters. I have very limited experience with R, but if it's not too much of a burden, I think it would be good to follow the roxygen documentation format. I've seen that format in other R scripts that our team has used and everyone following the same format makes it really easy to understand quickly. https://cran.r-project.org/web/packages/roxygen2/vignettes/roxygen2.html

2_process/src/filter_wqp_data.R
```diff
@@ -0,0 +1,32 @@
+filter_wqp_data <- function(data,params_select,units_select,select_wqp_vars,omit_wqp_events,fileout,exclude_tidal=TRUE){
+  
+  # params are values to keep from the param column of the harmonized wqp data
```
jsadler2, line 3: same comment as above re the format of documentation.

3_visualize/src/report-do-inventory.Rmd
```diff
@@ -0,0 +1,186 @@
+---
+title: "DRB dissolved oxygen data availability"
```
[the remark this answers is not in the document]
jsadler2: It's just so pretty that it'd be a shame not to be able to see it easily!

1_fetch/src/download_DO_data.R
```diff
@@ -0,0 +1,30 @@
...
+  write_csv(daily_data_out, file = fileout)
+  
+  return(fileout)
+  
+}
+
+
+download_cont_DO_data <- function(site_list,pcode,time_zone){
+  
+  cont_DO_data_ls <- lapply(site_list,function(x){
```
jsadler2, line 20: Can you just pass in a list of all the sites at once or do you have to do the `lapply` and then later `rbind`?

jsadler2 (commented):
Looks good, @lekoenig. Thanks for getting this rolling! I have a few detailed, in-line questions in comments. I also have some big picture questions: what kind of data are you actually getting? Am I right that you are just getting everything (as in you are not aggregating temporally at all?). Also, for the DV, are you just getting mean? If we also want min and max, what does that entail and do you have any ideas of how to best format those different "target variables" (min, mean, max)?

1_fetch.R
```diff
@@ -0,0 +1,41 @@
+source("1_fetch/src/find_nwis_sites.R")
+source("1_fetch/src/download_DO_data.R")
+
+p1_targets_list <- list(
+  # Load harmonized WQP data product
+  tar_target(
+    p1_wqp_data,
+    readRDS(file = "1_fetch/in/DRB.WQdata.rds")),
```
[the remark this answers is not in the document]
lekoenig: Sure, good question! `DRB.WQdata.rds` is a copy of the multiscale surface water quality data for the DRB (Shoda et al. 2019, data release [here](https://www.sciencebase.gov/catalog/item/5d6fa631e4b0c4f70cf92e9e)). The Shoda et al. dataset represents a QA/QC'ed version of the data available on the Water Quality Portal through early 2019. Ideally, I would pull this in programmatically with `sbtools`, but added this file to an `/in/` directory to get around issues with ScienceBase going down, at least temporarily.

1_fetch/src/download_DO_data.R
```diff
@@ -0,0 +1,30 @@
+download_daily_mean_DO_data <- function(site_list,pcode,stat_cd,fileout){
+  
+  daily_DO_data_ls <- lapply(site_list,function(x){
+    readNWISdv(siteNumbers = x,parameterCd=pcode,statCd=stat_cd,startDate = "",endDate = "") %>%
+      renameNWISColumns()
```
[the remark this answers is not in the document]
lekoenig: `readNWISdv()` comes from the dataRetrieval package and downloads the available daily data for a given site and parameter code(s). Would it help clarify the code chunk if I explicitly referenced dataRetrieval here, e.g. `dataRetrieval::readNWISdv(...)`? I'm not sure if USGS-R has style guidelines for this sort of thing.

3_visualize/src/report-do-inventory.Rmd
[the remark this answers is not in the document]
lekoenig: Sure, we could add and commit the rendered reports so that they're easily referenced as intermediate outputs/visual checks.

1_fetch/src/download_DO_data.R
lekoenig: Yep, you can pass in a list of all the sites at once and that would be more efficient here. I sometimes opt for the lapply-rbind steps if additional processing or variable name mungeing is needed on the downloaded data for individual sites, but that isn't happening here so I'll make the change to instead pass a list of sites to dataRetrieval. Thanks for pointing this out!

lekoenig:
> Looks good, @lekoenig. Thanks for getting this rolling! I have a few detailed, in-line questions in comments. I also have some big picture questions: what kind of data are you actually getting? Am I right that you are just getting everything (as in you are not aggregating temporally at all?). Also, for the DV, are you just getting mean? If we also want min and max, what does that entail and do you have any ideas of how to best format those different "target variables" (min, mean, max)?

Thanks, @jsadler2! A goal of this PR is to initiate these kinds of discussions about what input data we want to target to suit our modeling needs, and what we want that data to look like, so I appreciate your input!

To start, this PR aims to retrieve observations of [DO] from NWIS continuous sensors (preferred, where available), NWIS daily summaries (for NWIS sites that do not report instantaneous data; for now, just grab the mean), and discrete/snapshot samples available from the Water Quality Portal. So there's no temporal aggregation, etc. _yet_ since here I was focused on getting an inventory of the number of sites and observation-days. 

We can definitely summarize and output the NWIS data as daily min/mean/max for all sites where that information is reported or can be estimated from the high-frequency time series. I envision replacing the multiple targets for continuous, daily, and wqp data with one target that contains min/mean/max [DO] *as available* across all sites and days. 

I'd appreciate further input on the discrete samples. These samples have much greater spatial coverage but aren't always sampled repeatedly, so they represent a single snapshot measurement on the day the sample was taken. Can these data still be useful for modeling? Perhaps even in static model development to inform relevant features? Or to inform a model of mean daily [DO], which might itself contain information relevant to expected min/max? Pinging @galengorski here, too.

1_fetch.R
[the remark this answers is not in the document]
jsadler2: ah gotcha. thanks for that explanation. 27 MB is getting up into the range where I think we should think carefully if we want to "host" this in our github repo. I don't love the idea of totally relying on SB, but I since we have to rely on SB for other parts of this project that will have much bigger files (I.e., the met drivers that will be ~1 GB), I think it's best if we rely on it here too.

3_visualize/src/report-do-inventory.Rmd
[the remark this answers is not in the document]
jsadler2: Oooh. Can you try this? https://stackoverflow.com/a/39816334/5811195

GH wasn't rendering the HTML, but I think you can export to just `.md` using above and then it will render on GH.

1_fetch.R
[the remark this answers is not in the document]
lekoenig: I agree in terms of storage size and it's a good point about relying on the met drivers that are on SB. I'll go ahead and bring this in with ScienceBase. The data shouldn't need frequent rebuilds and I can try to think about other ways to build in some fault tolerance for when SB goes down.

1_fetch/src/download_DO_data.R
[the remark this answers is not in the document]
lekoenig: `renameNWISColumns` is also a helper function from dataRetrieval, so similarly, I could make that reference explicit if it would help with code readability.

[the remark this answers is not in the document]
jsadler2: Ah. Okay. I don't think there is a USGS-R style guideline for this. I guess it gets back to a fundamental R thing that leaves ambiguity when it comes to the source of a given function. I'll defer to you (and @galengorski, if you have an opinion) since I am a less frequent R user, whether to specify the source (e.g., `dataRetrieval::renameNWISColumns`) or leave it as is since it seems like R folks wouldn't be phased by this.

jsadler2:
> I'd appreciate further input on the discrete samples. These samples have much greater spatial coverage but aren't always sampled repeatedly, so they represent a single snapshot measurement on the day the sample was taken. Can these data still be useful for modeling? Perhaps even in static model development to inform relevant features?

I think it's a good idea to pull the discrete samples if it's not much more work. It's still an unknown how useful these would be to the model, but I'm guess these could be pretty useful for informing the spatial distribution (assuming there is some temporal overlap)

galengorski:
I agree that the discrete data might have use later on in the modeling effort. I can imagine aggregating discrete samples in similar areas to fine tune a model to make predictions in an area where don't have great coverage from the high resolution data. I also think that from a network design perspective a common question might be, "Where should we invest in installing a high frequency sensor, and where can we get away with making discrete measurements?" I wonder if we might be able to shine a light on that by strategically using the discrete data with the model.

1_fetch.R
```diff
@@ -0,0 +1,41 @@
...
+  tar_target(
+    p1_DO_sites_nwis,
+    get_drb_sites(drb_huc8s,DO_pcodes,site_tp_select)),
+  tar_target(
+    p1_daily_DO_csv,
```
jsadler2, line 13: I think it'd be more readable to break this out into it's own function (maybe `get_daily_do_sites` and it's own target).

1_fetch.R
```diff
@@ -0,0 +1,41 @@
...
+        } else {
+          data_type_cd == "uv"
+        }) %>%
+        ungroup() %>%
+        filter(data_type_cd == "dv")
+      # Download and save daily DO data
+      download_daily_mean_DO_data(site_list = unique(DO_daily_sites$site_no),pcode=DO_pcodes,stat_cd="00003",fileout="1_fetch/out/DRB_daily_DO_data.csv")
+    }),
+  tar_target(
+    p1_cont_DO_data_ls,
```
jsadler2, line 30: Is this target used anywhere?

jsadler2:
I was looking at this some more today. Are the diagonal white stripes the weekends?
image

lekoenig:
> I was looking at this some more today. Are the diagonal white stripes the weekends? image

I think so! Reminds me of cellular automata.

1_fetch/src/download_DO_data.R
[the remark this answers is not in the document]
lekoenig: I'm hoping that adding better function documentation as discussed below will help address issues like this (so thanks for pointing it out!), however I've also decided to adopt source notation (e.g. `dataRetrieval::renameNWISColumns`) within functions to make external function definitions more clear.

1_fetch/src/find_nwis_sites.R
lekoenig: I actually didn't know this, so that's a great tip! However, I noticed this sentence from the documentation:

> If you request a major site type (ex: &siteType=ST) you will get all sub-site types of the same major type as well (in this case, ST-CA, ST-DCH and ST-TS)

We request stream samples ("ST") but we don't want tidal samples for this work ("ST-TS"). Adding "ST" to siteType in the service call would then still require a filtering step to remove tidal sites, so I've decided to keep this code chunk as is and filter for site types within the dataRetrieval output.

1_fetch/src/find_nwis_sites.R
lekoenig: In the code edits within this PR I've added roxygen-style documentation to each of the functions. Thanks for this great suggestion that will help all of us read and interpret the code!

1_fetch.R
lekoenig: I agree that this step - to filter and identify the continuous NWIS sites for which there is no instantaneous (sub-daily) data reported but there are daily statistics - wasn't very clear/readable, especially given the original placement within the `tar_target()` call. To make `1_fetch.R` more readable, I've created a new target `p1_nwis_sites_daily` that represents just the daily sites, which gets called in `get_daily_nwis_data()` to download the data for each nwis "dv" site.

1_fetch/src/download_DO_data.R
lekoenig: One disadvantage of both the `lapply` way I had originally implemented and passing a list of all the sites to dataRetrieval is that _all_ of the data would be re-downloaded, for example, if just one site in the upstream site list changed. I learned some very useful tips and tricks from the last pipelines training course, so I'll use targets and dynamic branching for this sort of split-apply-combine workflow moving forward. I've gone ahead and implemented this sort of step with the daily data download and the `p1_daily_data` target. Now, if one site changes, we will only re-download the data for that one site rather than all sites. I'll use this same idea for the continuous data download, too.

3_visualize/src/report-do-inventory.Rmd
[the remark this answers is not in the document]
lekoenig: So this _seems_ like it should be trivial to do, but Rmarkdown makes working with subdirectories a pain. I tested this out a bit without much luck - even after exporting to .md the report isn't rendering correctly on GH.

I don't want to hold up this merge, so I suggest reverting to the USGS-R style of not committing /out files for now. I can open an issue to dig into this further depending on how we want to share visuals without necessarily building the pipeline locally. Sound OK, @jsadler2?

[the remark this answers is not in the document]
galengorski: @lekoenig I was messing around with this also,[ for another project ](https://github.com/galengorski/municipal-water-loss-ml/tree/main/1_data_inventory) and I was running into issues. I ended up putting a link to an rpub website in the repo and hosting the markdown file there. Not a great solution, but the markdown file is viewable without having to recompile.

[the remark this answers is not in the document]
lekoenig: Thanks, @galengorski - it seems like it should be so easy! I messed around with a few options and wasn't satisfied with anything I was coming up with. I like your idea to host on an rpub website at least in the interim. I'll see if that can be integrated within the targets pipeline.

lekoenig:
I suggest we can go ahead and merge this PR, which represents the initial site and data inventory. The pipeline currently builds a table called "DB_DO_sitelist.csv" which represents the site list we need for matching point observations to stream segments (issue #16). 

The next step is to output the combined DO data that can be used for modeling:

> I envision replacing the multiple targets for continuous, daily, and wqp [discrete] data with one target that contains min/mean/max [DO] as available across all sites and days.

3_visualize/src/report-do-inventory.Rmd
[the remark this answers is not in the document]
jsadler2: > I don't want to hold up this merge, so I suggest reverting to the USGS-R style of not committing /out files for now.

Sounds good to me. I do like the idea of figuring out a solution for sharing results (that RPub is cool!) but seems like outside of this PR would be best.

1_fetch/src/get_inst_nwis_data.R
```diff
@@ -0,0 +1,39 @@
+get_inst_nwis_data <- function(site_info,parameter) {
+  #' 
+  #' @description Function to download NWIS daily data
```
jsadler2, line 3:
```suggestion
  #' @description Function to download NWIS instantaneous data
```

_targets.R
```diff
@@ -0,0 +1,49 @@
...
+units_select = c("mg/l")
+
+# Define hydrologic event types in harmonized WQP data to exclude
+omit_wqp_events <- c("Spill","Volcanic action")
+
+# Define USGS parameter codes
+pcode_select <- c("00300") 
+
+# Define minor HUCs (hydrologic unit codes) that make up the DRB
+drb_huc8s <- c("02040101","02040102","02040104","02040103","02040106","02040105",
```
jsadler2, line 38: Maybe here we can just do the subset of the lower DRB (#13)?

jsadler2 (commented):
These edits look good, @lekoenig! Just a couple tiny suggestions that I noted inline.

Also, one question:

You said
>I envision replacing the multiple targets for continuous, daily, and wqp [discrete] data with one target that contains min/mean/max [DO] as available across all sites and days.

I'm already seeing the `p1_daily_data` and `p1_inst_data` already in this repo. So are you thinking the combination of those will be in another PR just dedicated to that?

_targets.R
lekoenig: Sure, we could subset the pipeline to only include the lower DRB. I added a comment to issue #13 earlier because I wasn't sure if the more extensive dataset would be helpful to retain for modeling.

jsadler2: I think we could go either way. My opinion is that we should make the decision to bring in outside data later on and start with only data from our study area.

jsadler2:
One other thing I think we should resolve before this merge and that is the inclusion of the `DRB.WQdata.rds` file in the repo history. Even though you removed it in a later commit, because it was earlier committed and pushed, it is part of the repo history and will therefore always be downloaded whenever the repo is cloned. I think the easiest way to remove it from the repo history is though the BFG tool (https://rtyley.github.io/bfg-repo-cleaner/). To run the BFG tool, you need Java installed and then it's pretty straightforward following the documentation. I can do that on my end if you like, @lekoenig - just let me know.

_targets.R
lekoenig: OK, sounds good! I've gone ahead and subset the data to only include the lower DRB

Issue #9, opened 2021-11-02
train/val/test splits


jsadler2:
We will need to come up with splits for train/validation/test splits.

galengorski:
Did stream temperature use a cross validation approach for training? We'll also need to think about this for hyperparameter tuning

jsadler2:
they held out some "validation" data. I don't know if they did any "cross-validation". 

And yes. we will definitely need to think about this for hyperparameter tuning.

jsadler2:
The temp project did quite a bit of work when they decided on their split: https://github.com/USGS-R/delaware-model-prep/issues/93

jsadler2:
I made some figures that show some overall data availability over time and by site:

### Daily sites (daily mean DO)
![daily_daily_means](https://user-images.githubusercontent.com/6943441/144119461-0d1300d6-1f5b-4c5b-8ad5-c15b1900819e.jpg)

### Instantaneous sites (daily mean DO)
![inst_daily_means](https://user-images.githubusercontent.com/6943441/144119578-76bd19e9-8e72-4e46-9ebb-210c6c7f29fd.jpg)

I also made a figure of the DOY averages across the sites:
![doy_means](https://user-images.githubusercontent.com/6943441/144119626-509097df-3617-4710-bf41-c89a875d4875.jpg)

jsadler2:
And here's a map of the points with the most well-observed points labeled:
![image](https://user-images.githubusercontent.com/6943441/144123680-f02c044a-98c6-4ede-82fe-321e642179b7.png)

jsadler2:
There's basically 5 sites with consistent "daily" data: 
![image](https://user-images.githubusercontent.com/6943441/144128360-39796db8-9607-48f8-bce7-73232bfb753d.png)
![image](https://user-images.githubusercontent.com/6943441/144128493-b95a96b0-dbeb-4502-b686-5682c91ee102.png)

jsadler2:
###  filtering to just the most well-observed instantaneous sites
And the 12 sites that have 300+ (arbitrary threshold) day-observations from the instantaneous record:
![image](https://user-images.githubusercontent.com/6943441/144129585-5a71cdec-5f35-49e1-99bd-b4408368465c.png)
![image](https://user-images.githubusercontent.com/6943441/144129643-369449c1-66c3-4284-8213-395c9e062bd7.png)

lekoenig:
[...]

PR #24, opened 2021-12-03
writing DO csvs and making summary plots

jsadler2:
- Writing out csvs for instantaneous and daily DO data. This is done for convenience so I can read the data into Python.
- Code that makes plots in #9

### DIFF
--- a/1_fetch/src/write_data.R
+++ b/1_fetch/src/write_data.R
@@ -1,0 +1,7 @@
+library(readr)
+
+
+write_to_csv <- function(data, outfile){
+  write_csv(data, file = outfile)
+  return(outfile)
+}
--- a/1_fetch/src/write_data.R
+++ b/1_fetch/src/write_data.R
@@ -1,5 +1,3 @@
-library(readr)
-
 
 write_to_csv <- function(data, outfile){
   write_csv(data, file = outfile)
--- a/3_visualize/src/do_overview_plots.py
+++ b/3_visualize/src/do_overview_plots.py
@@ -1,0 +1,55 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.11.2
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
+import pandas as pd
+import matplotlib.pyplot as plt
+
+df_inst = pd.read_csv('1_fetch/out/inst_do_data.csv', parse_dates=['dateTime'], infer_datetime_format=True, dtype={'site_no':str})
+
+df_inst = df_inst.drop_duplicates(subset=['site_no', 'dateTime'])
+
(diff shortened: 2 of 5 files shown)


1_fetch/src/write_data.R
```diff
@@ -0,0 +1,7 @@
+library(readr)
```
lekoenig, line 1: I think the code should run without the explicit call to `readr` here since the tidyverse collection of packages is included in `tar_option_set(...)` within the `_targets.R` file.

1_fetch/src/write_data.R
```diff
@@ -0,0 +1,7 @@
...
+
+
+write_to_csv <- function(data, outfile){
+  write_csv(data, file = outfile)
+  return(outfile)
+}
```
lekoenig, line 7: I wondered whether this function is so small that it doesn't warrant a separate function call. However, I think the alternative would be to code the `write_csv` function within each target since `targets` is looking for the outfile path:

    tar_target(p1_daily_data_csv,  
    {write_to_csv <- function(data,outfile){  
    readr::write_csv(data,outfile)  
    return(outfile)}   
    write_to_csv(p1_daily_data,"1_fetch/out/daily_do_data.csv")  
    },  
    format = "file")

Since this would have to be repeated for p1_inst_data_csv, I think the separate `write_to_csv` file is more readable and can be repeated for the daily and inst datasets.

3_visualize/src/do_overview_plots.py
```diff
@@ -0,0 +1,134 @@
...
+    color="steelblue",
+    layout=(6, 2),
+    alpha=0.6,
+)
+for i, site_id in enumerate(df_inst_means_filt.columns):
+    ax = axs.flatten()[i]
+    ax.legend(loc="upper left")
+
+fig = plt.gcf()
+fig.suptitle("daily minimum DO (mg/l) at daily sites", y=0.99)
```
lekoenig, line 132: Should this title read "daily mean DO (mg/l) at inst sites"?

jsadler2: oh yep! good catch.

1_fetch/src/write_data.R
jsadler2: Yeah that was exactly the thought process I went through too. Glad to know I wasn't too off-base. I think it's just enough of a difference from the `readr` function that it would kind of clog up the pipeline file.

1_fetch/src/write_data.R
jsadler2: Oh. Cool! I did not realize that.

jsadler2:
Hi @lekoenig - thanks for the review. I think I addressed your comments.

I also wanted to point out for future reference that I had the idea to add the python plotting file as a dependency to the plotting target: https://github.com/USGS-R/drb-do-ml/pull/24/commits/f2f0ab7f8f0d213a85cf61cef7b1fa04fd0e1087. This way, targets knows to rebuild if I change the .py file (like if I mislabel the an axis label 😄).

PR #25, opened 2021-12-08
aggregating instantaneous data to daily summaries

jsadler2:
The PR 
1. aggregates the instantaneous DO data to daily min/mean/max values and then 
2. combines them with the Daily data. 
 
It's a pretty simple implementation and there are likely some nuances that I didn't consider that I'm hoping you may catch, @lekoenig.


jsadler2:
superseded by #35

PR #26, opened 2021-12-10
16 point to reach

### LINKED ISSUE #16: match point observations to stream segments
we will need to match to the point observations to the stream segments. Once again, the temperature project has blazed the trail here, so we can look to and hopefully be able to reuse their code.

jsadler2:
Implementing stripped down version of David Watkin's point to reach mapping (https://code.usgs.gov/wwatkins/national-site-reach/-/blob/master/R/match_sites_reaches.R). 

The "stripping down" part happens in this commit: https://github.com/jsadler2/drb-do-ml/commit/c6e76929a5793e571cf59956e20fbb90b2dd6faa. From what I can gather (without having run the original code) what I stripped out was two main things:
1. I made it specific to our use case

- (`segidnat`) as the seg_ids
- (`lat`, `lon`) as the geometry columns 
- took out the bit about diverging reaches since those are already filtered out of the Geofabric

2. I removed the functionality of calculating distance from segment endpoints and associating sites with upstream ones if they are closer to the upstream outlet.

- It was a little tricky for me to figure out how to make this functionality (which was spread across several functions and files in David's repo) work for our case. I'm not suggesting that we shouldn't ever have this functionality, but I also don't know if we need it right away. That and given the complexity of adapting it to our use case, I'd rather have this as a first cut and then, if/when we decide we need that additional information, we can figure out how to work it in.

Also note that I'm suggesting that we just let the reaches shapefile live within the repo instead of asking targets to build it by downloading from science base:
https://github.com/jsadler2/drb-do-ml/blob/a94bbb39300de505f971851f4b1fcf859115d1f5/1_fetch.R#L67
```R
  tar_target(
    p1_reaches_shp_zip,
    # [Jeff] I downloaded this manually from science base: 
    # https://www.sciencebase.gov/catalog/item/5f6a285d82ce38aaa244912e
    # Because it's a shapefile, it's not easily downloaded using sbtools
    # like other files are (see https://github.com/USGS-R/sbtools/issues/277).
    # Because of that and since it's small (<700 Kb) I figured it'd be fine to
    # just include in the repo and have it loosely referenced to the sb item ^
    "1_fetch/in/study_stream_reaches.zip",
    format = "file"
  ),
```

Also, I realized that the first two commits of aggregating instantaneous data to daily are included in another, PR #25. Whoops. Let's try to focus this conversation on just the point to reach matching.

closes #16


jsadler2:
see parallel PR in the inland-salinity repo: https://github.com/USGS-R/drb-inland-salinity-ml/pull/27

jsadler2:
superseded by #35

Issue #31, opened 2021-12-17
unusual data from site `01482537`


jsadler2:
I'm getting unusual data from `readNWISuv` for site `01482537` (https://waterdata.usgs.gov/nj/nwis/inventory/?site_no=01482537)

```R
head(dataRetrieval::readNWISuv(siteNumbers = "01482537",parameterCd="00300",startDate = "",endDate = "",tz = "UTC") %>% 
    dataRetrieval::renameNWISColumns(p00300="Value",p00095="Value"))
  agency_cd  site_no            dateTime at.0.5.ft.depth_Value_Inst
1      USGS 01482537 2008-08-06 17:45:00                       13.5
2      USGS 01482537 2008-08-06 18:00:00                       13.3
3      USGS 01482537 2008-08-06 18:15:00                       14.7
4      USGS 01482537 2008-08-06 18:30:00                       14.7
5      USGS 01482537 2008-08-06 18:45:00                       14.8
6      USGS 01482537 2008-08-06 19:00:00                       14.5
  at.0.5.ft.depth_Value_Inst_cd at.2.0.ft.depth_Value_Inst
1                             A                         NA
2                             A                        9.6
3                             A                       10.5
4                             A                       10.3
5                             A                       10.6
6                             A                       11.6
  at.2.0.ft.depth_Value_Inst_cd tz_cd
1                          <NA>   UTC
2                             A   UTC
3                             A   UTC
4                             A   UTC
5                             A   UTC
6                             A   UTC
```

I think it has something to do with the `at.0.5.ft.depth_Value_Inst` column names and that throws off some downstream processes.

For reference. When I do a different site (`01472104`) I get this:
```R
head(dataRetrieval::readNWISuv(siteNumbers = "01472104",parameterCd="00300",startDate = "",endDate = "",tz = "UTC") %>%
     dataRetrieval::renameNWISColumns(p00300="Value",p00095="Value"))
  agency_cd  site_no            dateTime Value_Inst Value_Inst_cd tz_cd
1      USGS 01472104 2007-10-01 05:00:00        6.9             A   UTC
2      USGS 01472104 2007-10-01 06:00:00        6.7             A   UTC
3      USGS 01472104 2007-10-01 07:00:00        6.6             A   UTC
4      USGS 01472104 2007-10-01 08:00:00        6.5             A   UTC
5      USGS 01472104 2007-10-01 09:00:00        6.5             A   UTC
[...]

Issue #29, opened 2021-12-17
set earliest date for data retrieval


jsadler2:
similar to the inland salinity project, I would like to have an `earliest_date` argument so we aren't pulling data from before `1979-10-01`

PR #32, opened 2021-12-20
Incorporate updates to data pipeline made in drb-inland-salinity

lekoenig:
This PR incorporates updates to the data pipeline that resulted from discussions in the inland-salinity repo (namely inland-salinity [PR 19](https://github.com/USGS-R/drb-inland-salinity-ml/pull/19) and inland-salinity [PR 31](https://github.com/USGS-R/drb-inland-salinity-ml/pull/31)). These changes include:

- Add `earliest_date` to the data retrieval calls so that we are not pulling data before ~1980 when the met data begins (resolves issue #29)
- Add special handling for site 01482537, which was recently added to NWIS. See discussion in [Teams](https://teams.microsoft.com/l/message/19:264c78fefa624bcfaf46ac10b9305f5e@thread.skype/1639492810490?tenantId=0693b5ba-4b18-4d7b-9341-f32f400a5494&groupId=2651eb2e-8cf1-4caa-a084-0a11facc1d36&parentMessageId=1639492810490&teamName=GS-WMA%20IIDD%20Staff&channelName=Function%20-%20Data%20Pipelines&createdTime=1639492810490) for further details. (resolves issue #31)
- Add log/ind files to the repo to better track data changes through time.
- Update the site list to find the CRS used for the WQP data and project to a single, user-specified datum


2_process.R
```diff
@@ -12,8 +12,9 @@ p2_targets_list <- list(
   
   # Subset harmonized WQP data to lower DRB
   tar_target(
-    p2_filtered_wqp_data_subset,
-    subset_wqp_sites(p1_lowerdrb_boundary,p2_filtered_wqp_data)),
+    p2_wqp_data_subset_csv,
```
jsadler2, line 15: What do you think about making this an object targets instead of file targets? I think it's quite a bit more convenient to do that. Then we don't have to deal with the file io. At first I was thinking that I'd be accessing these via Python, but I'm finding it more convenient to stay in R until even further down the road. We could probably take out the `p1_daily_data_csv` and `p1_inst_data_csv` targets too.

2_process.R
```diff
@@ -24,7 +25,11 @@ p2_targets_list <- list(
   # Create a list of unique site locations containing DO data  
   tar_target(
     p2_site_list_csv,
-    create_site_list(p2_filtered_wqp_data_subset,p1_lowerdrb_boundary,p1_nwis_sites,p1_daily_data,p2_inst_data_hourly,fileout = "2_process/out/DRB_DO_sitelist.csv"),
+    {
```
jsadler2, line 28: Same comment here as above. I'm thinking we should just make our lives simpler by doing an object target instead of a file one here.

3_visualize.R
```diff
@@ -16,6 +19,44 @@ p3_targets_list <- list(
              format = "file",
              pattern = map(p2_inst_data_hourly)),
   
+  # Create and save log file containing data availability summary
+  tar_target(
+    p3_sitelist_summary_csv,
+    summarize_site_list(p2_site_list_csv,p1_daily_data,p1_inst_data,fileout = "3_visualize/log/sitelist_summary.csv"),
+    format = "file"),
```
jsadler2, line 26: To me this seems like this log file should be in the same directory as the actual file, so `2_process/log/`. I wouldn't think to look in the `visualization` directory for a log file that describes a target build in `2_process.R`. What do you think of the idea of moving it to `2_process/log`

3_visualize.R
```diff
@@ -16,6 +19,44 @@ p3_targets_list <- list(
...
+  tar_target(
+    p3_inst_timeseries_ind_csv,
+    command = save_target_ind_files("3_visualize/log/inst_timeseries_ind.csv",names(p3_hourly_timeseries_png)),
+    format = "file"),
+  
+  # Create and save indicator file for WQP data
+  tar_target(
+    p3_wqp_ind_csv,
+    command = save_target_ind_files("3_visualize/log/wqp_data_ind.csv","p2_wqp_data_subset_csv"),
+    format = "file"),
```
jsadler2, line 44: similar to above, I think it would be more intuitive to have this log file in `2_process/log`

3_visualize.R
```diff
@@ -16,6 +19,44 @@ p3_targets_list <- list(
...
+    command = target_summary_stats(p1_daily_data,"Value","3_visualize/log/daily_timeseries_summary.csv"),
+    format = "file"
+  ),
+  
+  # Create and save summary log file for NWIS instantaneous data
+  tar_target(
+    p3_inst_timeseries_summary_csv,
+    command = target_summary_stats(p1_inst_data,"Value_Inst","3_visualize/log/inst_timeseries_summary.csv"),
+    format = "file"
+  ),
```
jsadler2, line 58: And these ones should go in `1_fetch/log` IMO

3_visualize/src/summarize_site_list.R
```diff
@@ -0,0 +1,46 @@
+summarize_site_list <- function(site_list_path,nwis_daily_data,nwis_inst_data,fileout){
```
jsadler2, line 1: If we are moving the `sitelist_summary_csv` target to `2_process.R` this file should then move to `2_process/src/`

3_visualize/src/summarize_timeseries.R
```diff
@@ -0,0 +1,49 @@
...
+    select(tar_name = name, filepath = path, hash = data) %>%
+    mutate(filepath = unlist(filepath))
+
+  # Save indicator table
+  readr::write_csv(ind_tbl, fileout)
+  return(fileout)
+}
+
+
+target_summary_stats <- function(df,ValueVar,fileout){
```
jsadler2, line 21: This function would go in `1_fetch/src`

_targets.R
```diff
@@ -10,6 +10,7 @@ source("3_visualize.R")
 dir.create("1_fetch/out/", showWarnings = FALSE)
 dir.create("2_process/out/", showWarnings = FALSE)
 dir.create("3_visualize/out/", showWarnings = FALSE)
+dir.create("3_visualize/log/", showWarnings = FALSE)
```
jsadler2, line 13: we would need to add 
```suggestion
dir.create("1_visualize/log/", showWarnings = FALSE)
dir.create("2_visualize/log/", showWarnings = FALSE)
dir.create("3_visualize/log/", showWarnings = FALSE)
```

jsadler2 (commented):
Looks good overall, @lekoenig. Thanks for submitting this! I made a few comments and questions below.

2_process/src/create_site_list.R
```diff
@@ -1,14 +1,14 @@
...
   #' @description Function to create one site list that contains unique site locations for modeling
   #'
   #' @param wqp_data data frame containing filtered data from the DRB harmonized water quality dataset for discrete samples.
   #' wqp_data must contain the following columns: c("LatitudeMeasure","LongitudeMeasure","ActivityStartDate","MonitoringLocationIdentifier",
   #' "MonitoringLocationName","OrganizationIdentifier")
-  #' @param ws_boundary an sf object representing the area of interest AOI (used to retrieve original data CRS information for WQP sites)
   #' @param nwis_sites data frame containing all NWIS continuous sites identified
   #' @param nwis_daily_data data frame containing daily data for all NWIS daily sites
   #' @param nwis_inst_data data frame containing instantaneous data for all NWIS instantaneous sites
+  #' @param crs_out character string indicating desired crs. Defaults to "NAD83", other options include "WGS84".
```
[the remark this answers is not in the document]
lekoenig: Oh good call, thanks!

jsadler2:
One thing that I've been thinking about lately: there are some parts of the pipeline that we aren't using currently and may not be using in the near future. These include
* The hourly summaries
* The hourly/daily plots of each station

I'm really glad we have that code in the repo so we can use it if/when we need to, but I'm thinking for now, we can just take out those targets from the pipeline so we aren't building targets we aren't currently needing. And then down the line, we can always add those back in.  

Even tho they are nice to have for an overview, I think that we can just use those functions as needed instead of plotting for every station automatically.

3_visualize.R
lekoenig: I initially put them all in the same `3_visualize/log` directory because I thought it might make reviewing/referencing them easier if they were all in one place. But I definitely see your point (and have wondered the same thing myself). If it'd be more intuitive, I can re-arrange the log files into phase-specific log folders.

jsadler2:
I'm actually kind of thinking we maybe should take out the Python plots too 😄. They kind of take a while to build and then without them we can get rid of those csv targets in 1_fetch.R. They were nice to have to get a feel for the data, but I'm not sure we need to keep them around. I could be pursuaded otherwise tho about both the python targets and the plots from r.

Maybe we do a happy medium and produce one summary plot with ggplot (to avoid the python rigmarole)

3_visualize.R
jsadler2: Ah. I see. In the pipelines I've seen they are in the same "number" directory as the target itself: https://github.com/USGS-R/delaware-model-prep/tree/main/2_observations/out

2_process.R
lekoenig: > We could probably take out the p1_daily_data_csv and p1_inst_data_csv targets too.

Right now, `p1_daily_data_csv` and `p1_inst_data_csv` are called within `3_visualize/src/overview_plots.py` so if I deleted those targets the pipeline wouldn't build. I could try to emulate some of the figures in R if you'd prefer to get rid of these file targets (probably in a new issue).

3_visualize.R
lekoenig: OK, that makes sense - thanks for linking to that example. I've also created a separate log subdirectory based on an example repo in one of the targets trainings. Do you have any preferences for sticking with /log folders vs. writing files to /out as in delaware-model-prep?

2_process.R
jsadler2: Ha yeah. I actually just wrote this below :)
https://github.com/USGS-R/drb-do-ml/pull/32#issuecomment-998305511

>I'm actually kind of thinking we maybe should take out the Python plots too 😄. They kind of take a while to build and then without them we can get rid of those csv targets in 1_fetch.R. They were nice to have to get a feel for the data, but I'm not sure we need to keep them around. I could be pursuaded otherwise tho about both the python targets and the plots from r.

>Maybe we do a happy medium and produce one summary plot with ggplot (to avoid the python rigmarole)

jsadler2: So yeah. My vote would be for "one summary plot with ggplot". And I agree that this would be better in a separate issue/PR.

3_visualize.R
jsadler2: I'd be fine either way. Slight preference to a `log` sub-directory instead of `out` to keep things a little cleaner in `out`

lekoenig: OK, I agree with that too. I'll go ahead and separate out those files, but keep the log folders within individual phases.

2_process.R
lekoenig: OK, I am going to leave `p1_daily_data_csv` and `p1_inst_data_csv` for now (and `overview_plots.py`). I've opened a new issue (#33) for revisiting these summary plots using R (and the object targets as opposed to file targets).

2_process.R
lekoenig: I agree with you that making the (discrete data) wqp subset above into an objects file rather than a targets file might be easier to work with. I also assume that to prep the discrete data as modeling inputs, we will need at least one more downstream step to combine with the daily continuous values. So it probably makes more sense for `p2_filtered_wqp_data_subset` to be an object target. 

I'm not as sure on the site list - this table contains the lat/lon of the unique site id's and contains basic summaries/tallies of the available data. The csv does get read in within the Rmd file, but I guess I'm wondering whether this target is useful as a file format so that it can be accessed and viewed outside of the pipeline.

lekoenig:
> One thing that I've been thinking about lately: there are some parts of the pipeline that we aren't using currently and may not be using in the near future. These include
> 
> The hourly summaries
> The hourly/daily plots of each station

I've kept the functions for reference but removed the targets that create the hourly aggregated values as well as the hourly/daily plots (https://github.com/USGS-R/drb-do-ml/pull/32/commits/eb38f58b188c2d2247622f39be1027ae603ee9f6).

2_process.R
jsadler2: That's a good point that this could be quite useful outside the pipeline. Now that you say that, I'm realizing that I am saving the one with the reach_ids to csv in #35. So let's leave the `p2_site_list_csv` as a file target in the PR, but I think it gets superseded by the `p2_sites_w_segs_csv` target in #35

lekoenig: Ah, OK. That makes sense that this table is superseded with the target that contains the paired segs.

lekoenig:
I think that if we opt to leave `p2_site_list_csv` as a file target, I've addressed and committed all of the changes we've discussed here. Let me know if you see anything else before we merge.

jsadler2:
Looks great. Thanks, @lekoenig

jsadler2:
Feel free to merge whenever

PR #35, opened 2021-12-21
reach matching and daily agg

### LINKED ISSUE #16: match point observations to stream segments

### LINKED ISSUE #34: aggregate inst. data to daily summaries
need to aggregate inst. data to daily summaries

jsadler2:
This PR is adds two main functionalities. Both main parts were discussed in the inland-salinity repo so this should (hopefully) go pretty quick.
1. site-to-reach matching (closes #16) - see https://github.com/USGS-R/drb-inland-salinity-ml/pull/27 for relevant discussion
2. temporal aggregation of instantaneous to daily summaries (closes #34) - see https://github.com/USGS-R/drb-inland-salinity-ml/pull/26 for relevant discussion.


lekoenig:
Hey @jsadler2 do you mind pulling changes from `main` into `16-34-reach-matching-daily-agg` before I review?

jsadler2:
Just did a rebase onto main, @lekoenig, so it should be good now.

1_fetch.R
```diff
@@ -88,8 +88,35 @@ p1_targets_list <- list(
...
+  ),
   
-)
-
+  tar_target(
+    p1_reaches_shp_zip,
+    # [Jeff] I downloaded this manually from science base: 
+    # https://www.sciencebase.gov/catalog/item/5f6a285d82ce38aaa244912e
+    # Because it's a shapefile, it's not easily downloaded using sbtools
+    # like other files are (see https://github.com/USGS-R/sbtools/issues/277).
```
lekoenig, line 98: Thanks for pointing to this sbtools issue - it confirms an issue I've run into before. I wonder if there are any plans to add functionality for pulling down shp files

jsadler2: IDK if there are plan for adding that functionality, but I think Blodgett made the issue (https://github.com/USGS-R/sbtools/issues/277) and tagged it `enhancement` so that hopefully one day they will make it easier to do these downloads.

2_process/src/create_site_list.R
```diff
@@ -1,4 +1,4 @@
-create_site_list <- function(wqp_data,nwis_sites,nwis_daily_data,nwis_inst_data,hucs,crs_out="NAD83",fileout){
+create_site_list <- function(wqp_data,nwis_sites,nwis_daily_data,nwis_inst_data,hucs,crs_out="NAD83"){
```
lekoenig, line 1: We can delete the description for "@param fileout" below. I think this change to an object target makes sense since it's more or less replaced by `p2_sites_wi_segs`. I haven't tried running the code yet - I wonder if we'll need to make any changes to the .Rmd file for the pipeline to build?

2_process/src/match_sites_reaches.R
```diff
@@ -0,0 +1,65 @@
...
+  #'
+  #' @param reach_sf sf object of reach polylines with column "subsegid" and in WGS84
+  #' @param sites dataframe with columns "lat" "lon"
+  #' @param sites_crs the crs os the sites table (i.e., 4269 for NAD83)
+  #' @param max_matches the maximum number of segments that a point can match to
+  #' @param search_radius the maximum radius in same units as sf object
+  #' within which a segment will match (segments outside of the radius will not match)
+  #'
+  #' @value A data frame the same columns as the sites input dataframe with additional columns
+  #' of "segidnat", "subsegid" and "offset" where "offset" is the distance (in degrees) between the point
```
lekoenig, line 17: I'm curious whether you have ideas for how the `offset` information might be used downstream. I struggle to think in units of degrees, so I'm not sure how I would use that column.

2_process/src/match_sites_reaches.R
```diff
@@ -0,0 +1,65 @@
...
+  #' and matching flowline
+  
+  # set up NHDPlus fields used by get_flowline_index
+  # Note: that we are renaming the subsegid column to COMID because the nhdplusTools
```
lekoenig, line 21: Thanks for adding this documentation! I think it will help me to remember how we're using the nhdplusTools functions for the PRMS flowlines.

2_process/src/match_sites_reaches.R
```diff
@@ -0,0 +1,65 @@
...
+  # Note: the `ToMeas` and `FromMeas` are also required columns for the nhdplusTools 
+  # function. Since we are using our own reaches and not the nhd, these do not have 
+  # the same meaning as they would if we were using the nhd
+  reaches_nhd_fields <- reach_sf %>%
+    rename(COMID = subsegid) %>%
+    mutate(REACHCODE = COMID, ToMeas = 100, FromMeas = 100) %>%
+    st_as_sf()
+  
+  sites_sf <- sites %>% rowwise() %>%
+    filter(across(c(lon, lat), ~ !is.na(.x))) %>%
```
lekoenig, line 33: In `filter_wqp_data.R` we filter to only retain sites that have a lat/lon (otherwise we get errors when using spatial functions to build the site list). So I don't think this line is doing anything here, but we may as well keep it since it's not hurting anything and we may want to port this function elsewhere in the future.

2_process/src/match_sites_reaches.R
```diff
@@ -0,0 +1,65 @@
...
+    mutate(Shape = list(st_point(c(lon, lat), dim = "XY"))) %>%
+    st_as_sf() %>% st_set_crs(sites_crs) %>%
+    st_transform(st_crs(reaches_nhd_fields)) %>%
+    st_geometry()
+  message('matching flowlines with reaches...')
+  flowline_indices <- nhdplusTools::get_flowline_index(flines = reaches_nhd_fields,
+                                                       points = sites_sf,
+                                                       max_matches = max_matches,
+                                                       search_radius = search_radius) %>%
```
lekoenig, line 42: Is 0.1 the default `search_radius` in nhdplusTools? Effectively, this nearest neighbor function will search within that radius (~10 km, I think) and if any segments are returned, we just want the closest one (e.g. maybe that line is 5 m or 5 km away). One thing I hadn't considered before is that with this approach, we may be incorrectly matching samples taken in small, headwater streams to larger river segments, which are the only ones represented by the PRMS lines. Do you think that's true?

I'm not suggesting any changes to the code here, perhaps just something to be aware of (and without looking at the returned matches more closely, I'm not sure whether it's actually an issue that comes up or how often).

2_process.R
```diff
@@ -24,17 +25,58 @@ p2_targets_list <- list(
...
+    aggregate_data_to_daily(p1_inst_data,p1_daily_data, min_daily_coverage=0.5, output_tz="America/New_York")),
+
+  # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
+  tar_target(
+    p2_daily_combined,
+    bind_rows(p1_daily_data, p2_inst_data_daily)),
+  
   # Create a list of unique site locations containing DO data  
   tar_target(
     p2_site_list_csv,
```
lekoenig, line 40: change target name to `p2_site_list`

2_process.R
```diff
@@ -24,17 +25,58 @@ p2_targets_list <- list(
...
-                       hucs=drb_huc8s,crs_out="NAD83",fileout = "2_process/out/DRB_DO_sitelist.csv"),
-    format = "file"),
-  
+                       hucs=drb_huc8s,crs_out="NAD83")
+  ), 
+
   # Create and save log file containing data availability summary
   tar_target(
     p2_sitelist_summary_csv,
     summarize_site_list(p2_site_list_csv,p1_daily_data,p1_inst_data,fileout = "2_process/log/sitelist_summary.csv"),
```
lekoenig, line 48: replace `p2_site_list_csv` to `p2_site_list` to reflect target format change

2_process.R
```diff
@@ -24,17 +25,58 @@ p2_targets_list <- list(
...
-    format = "file")
+    format = "file"),
+
+  tar_target(
```
lekoenig, line 51: Would you mind adding some one-line comments above each `tar_target()` call to describe the target? Nothing fancy, but I've noticed these commented descriptors help me to quickly navigate the phase files without needing to closely examine each target.

2_process/src/match_sites_reaches.R
lekoenig: It looks like 50% of sites have offsets at or below 0.005 (~0.5 km, I think), but some can be higher, up to our max `search_radius`

offsets

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,49 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+    group_by(site_no, Date, agency_cd, Parameter) %>%
+    summarise(Value = mean(Value_Inst, na.rm=TRUE), 
+              Value_Min = min(Value_Inst, na.rm=TRUE), 
+              Value_Max = max(Value_Inst,na.rm=TRUE), 
+              na_count=sum(is.na(Value_Inst)), 
+              value_count=sum(!is.na(Value_Inst)),
+              Value_cd = first(Value_Inst_cd),
+              Value_Max_cd = first(Value_Inst_cd),
+              Value_Min_cd = first(Value_Inst_cd),
+              .groups="keep") %>%
```
lekoenig, line 96: When I try to build the pipeline, I get a warning message from this target that Inf values are being returned. Do you see that too? It looks like that's happening because site 01475548 contains at least one day (6/6/2021) where all 96 values are `NA`. So when we try to take the min/mean/max for that day, there are no values to compute those summaries. It looks like we still get the intended result because those days get filtered out of the `daily_values` df anyway (`percent_coverage` is 0). 

If you wanted to get rid of the warning message, I think you could alter this code chunk a bit and change the order of the filtering steps and the summarizing steps. That might look like the code below, although I didn't run/compare the output with `p2_inst_data_daily` 

```
 daily_values <- inst_data %>%
    filter(site_no %in% only_inst_data) %>%
    filter(site_no == only_inst_data[9]) %>%
    mutate(dateTime_local = lubridate::with_tz(dateTime,tzone=output_tz),
           Date = lubridate::date(dateTime_local)) %>%
    group_by(site_no, Date, agency_cd, Parameter) %>%
    mutate(na_count=sum(is.na(Value_Inst)), 
           value_count=sum(!is.na(Value_Inst)),
           percent_coverage=value_count/(value_count + na_count)) %>%
    filter(percent_coverage >= min_daily_coverage) %>%
    summarise(Value = mean(Value_Inst, na.rm=TRUE), 
              Value_Min = min(Value_Inst, na.rm=TRUE), 
              Value_Max = max(Value_Inst,na.rm=TRUE), 
              Value_cd = first(Value_Inst_cd),
              Value_Max_cd = first(Value_Inst_cd),
              Value_Min_cd = first(Value_Inst_cd),
              .groups="keep") 
```

2_process/src/create_site_list.R
lekoenig: Line 20 in the .Rmd file should be changed to:

`site_list_path <- tar_read(p2_sites_w_segs_csv)`

lekoenig (changes requested):
Thanks, @jsadler2! This looks good. I've requested a couple minor changes to add documentation and make edits so that the pipeline will build properly. 

I also added a thought I had about the pt to segs matching - I'm not requesting any changes to our code or approach, but I'd be interested in your thoughts.

2_process/src/match_sites_reaches.R
jsadler2: I hadn't thought of the headwater streams not being in the network that is being matched to. I think that's good to keep in mind. I'll make an issue for this so we don't forget. For this first model, I was planning on using just the most well-observed sites, so I'm guessing that those will be on a bit bigger streams (ie., I'm guessing that they will be correctly matching to our PRMS segments).

2_process/src/munge_inst_timeseries.R
jsadler2: Thanks for figuring that out. I think this is a good change.

jsadler2 (commented):
Thanks for the feedback, @lekoenig . I think I addressed all of your comments.

2_process/src/match_sites_reaches.R
lekoenig: Sounds like a good plan!

2_process/src/match_sites_reaches.R
lekoenig: OK, sounds good and I would agree with your assessment of the most observed sites. _If_ this is happening, perhaps it's just a matter of adjusting the search radius. As a first impression/ballpark, I would think a GPS point shouldn't plot more than 300-500m away from the intended stream. Thanks for making the new issue.

2_process.R
lekoenig: Thanks for adding those!

3_visualize/src/report-do-inventory.Rmd
```diff
@@ -17,8 +17,7 @@ options(scipen=999)
 DO_wqp_data <- tar_read(p2_filtered_wqp_data_subset)
 
 # Load full site list:
-site_list_path <- tar_read(p2_site_list_csv)
```
lekoenig, line 20: The leaflet map further down in this document calls `site_list_path` so in order for the pipeline to build, we either need to 

- adjust the `map_sites` function to take `site_list` instead of `site_list_csv` (maybe this is the better longer term solution; this function is in `3_visualize/map_sites.R), or
- revert lines 20-21 here (for site_list_path, could just replace with `p2_sites_w_segs_csv`, since that is the same site list but with the new segid variable)

lekoenig:
> Thanks for the feedback, @lekoenig . I think I addressed all of your comments.

I think there's just one lingering target name thing to address before we merge. See my comment above.

jsadler2:
oh boy. i guess that's what i get for not actually building the thing. thanks for your patience on this. 

I've now actually successfully built it and I think it's good to go

lekoenig:
> oh boy. i guess that's what i get for not actually building the thing. thanks for your patience on this.
> 
> I've now actually successfully built it and I think it's good to go

No worries, that's what code reviews are for! It looks good to me and I was able to successfully build the pipeline. I approve to go ahead and merge.

lekoenig (approved):

PR #37, opened 2021-12-28
Add target to fetch network attributes

lekoenig:
This PR addresses issue #17. Two targets are added to the pipeline:

1. A target to download the DRB network attributes from ScienceBase
2. A target to load the DRB network adjacency matrix from the downloaded attributes

PR #39, opened 2021-12-28
Add targets for met data and segment attributes

### LINKED ISSUE #11: gather sntemp/prms meteorological data
this should be pretty simple. just need to pull the file from the Temperature SB item: https://www.sciencebase.gov/catalog/item/5f6a289982ce38aaa2449135 [wrong link earlier]

### LINKED ISSUE #12: Gather sntemp/prms segment/catchment attributes
For a first crack, I think we should try the model with just the sntemp/prms segment/catchment data. Then we can add NHD/streamcat data later on. So we will need to "gather" this data. 

I'm not sure exactly how best to do this. It's part of the delaware_model_prep pipeline, but 
1. it's not fully integrated into that pipeline yet
2. i don't know the best way to leverage that. Maybe we just say: "this is where we got it" and leave it at that, trusting the other pipeline.

### LINKED ISSUE #38: Generic function for downloading sb data
It would be good to have a more generic function to download data from science base. I'm thinking that the arguments for this function will be `sb_id`, `filename`, `out_dir`.

jsadler2:
Adding targets for prms meteorological data and prms segment attribute data. The met data comes from science base. The segment attribute data is not currently published but built in the [delaware-model-prep](https://github.com/USGS-R/delaware-model-prep) pipeline. 

I also made a generic wrapper function for downloading files from science base. @lekoenig - if you are in agreement, I'd like to merge this before #37. Then we can use the new function `download_sb_file` for pulling the adjacency matrix.

closes #38
closes #11,
closes #12


1_fetch.R
```diff
@@ -118,6 +131,41 @@ p1_targets_list <- list(
   tar_target(
     p1_reaches_sf,
     st_read(p1_reaches_shp)
+  ),
+
+  # fetch prms met data
+  tar_target(
+    p1_prms_met_data_zip_file,
```
lekoenig, line 138: Above we named targets ending in "_zip" and "_csv" so if we wanted to save ourselves a few keystrokes we could omit the "_file" suffix here and in the prms met data target. This is a very minor comment and I don't have a preference on these target names.

1_fetch/src/fetch_sb_data.R
```diff
@@ -0,0 +1,19 @@
+download_sb_file <- function(sb_id, file_name, out_dir){
```
lekoenig, line 1: I like that having this wrapper function makes it easy to return the file path that is useful for `format="file"` (as opposed to calling `item_file_download` within the target itself, which probably is not as easy to read anyway).

1_fetch/src/fetch_sb_data.R
```diff
@@ -0,0 +1,19 @@
...
+  #' @param file_name string - the name of the file in the science base item to download
+  #' @param out_dir string - the directory where you want the file downloaded to
+  #'
+  #' @value string the out_path
+
+  out_path = file.path(out_dir, file_name)
+  # Get the data from ScienceBase:
+  sbtools::item_file_download(sb_id = sb_id,
+                              names = c(file_name),
+                              destinations = c(out_path),
```
lekoenig, line 15: I wondered why the `c()` was used here. I didn't realize (or had forgotten) that you could input a string vector to only return file names of interest. That might come in handy for downloading the network adjacency matrix and not the other network attribute files associated with the same data release and `sb_id`.

1_fetch.R
```diff
@@ -118,6 +131,41 @@ p1_targets_list <- list(
...
+
+  # read in prms met data
+  tar_target(
+    p1_prms_met_data,
+    read_csv(p1_prms_met_data_csv_file)
+  ),
+
+  # read in prms met data
+  # [Jeff] I'm including these in the "in" folder because they are unpublished
+  # They are built in the delaware_model_prep pipeline (1_network/out/seg_attr_drb.feather)
```
lekoenig, line 163: This makes good sense to me. Do you know what the reason was for using .feather files in `delaware-model-prep`? Faster read/write? Can .feather files be read into python, too?

1_fetch.R
```diff
@@ -118,6 +131,41 @@ p1_targets_list <- list(
...
+    unzip(zipfile=p1_prms_met_data_zip_file,exdir = dirname(p1_prms_met_data_zip_file),overwrite=TRUE)
+    file.path(dirname(p1_prms_met_data_zip_file), "sntemp_inputs_outputs_drb.csv")
+    },
+    format = "file"
+  ),
+
+  # read in prms met data
+  tar_target(
+    p1_prms_met_data,
+    read_csv(p1_prms_met_data_csv_file)
```
lekoenig, line 158: Can add `show_col_types = FALSE` as an argument to `read_csv` to suppress messages about interpreted column data types with pipeline build.

lekoenig (approved):
Looks good to me, @jsadler2! The pipeline builds locally for me as well. I've made a couple of minor comments but none are necessary changes. I approve to merge whenever you're ready.

1_fetch/src/fetch_sb_data.R
jsadler2: I put the `c()`  because in the [documentation ](https://rdrr.io/github/USGS-R/sbtools/man/item_file_download.html) it says it's a string _vector_. So I think you can pass multiple file names which is what I was thinking. But I tried without the `c()` and it worked. So I'll take those out.

lekoenig: @jsadler2 Just to clarify, I don't think you should necessarily take those out. I was just commenting that I didn't know you could do that (but sure enough, it's clear in the documentation) - good to know about!

1_fetch.R
jsadler2: Good question. I was the one that wrote all of this and now looking back, I don't think it was a good idea. I think csv would be much better. You can read feather in Python. I did it in feather b/c it is slightly smaller and because it preserves the data types, but now I think it would be better just to do csv for simplicity/transparency. That said, I'm not sure if/when it's worth switching that over.

jsadler2:
Thanks @lekoenig for the comments and helpful suggestions!

1_fetch.R
lekoenig: Thanks for the context! R has a file format (.rds) that preserves data types and I think has faster read/write, but I'm hesitant to use it sometimes because it doesn't seem portable to other languages - it's useful that feather preserves data types and can be read into Python too ❤️

Issue #48, opened 2022-01-31
Check correlation between temperature and DO predictions


jsadler2:
We are interested to see the correlation between model DO predictions and input air temperature in the Baseline model.

jsadler2:
![do_preds_vs_temp](https://user-images.githubusercontent.com/6943441/154337089-a31abc2d-c833-4ed2-9275-28dbd5650c1c.png)

Here are plots of DO (predicted and observed) vs stream temperature. 

### The lower sites
The lower row of sites are doing quite well I'd say. I'm looking for places where the predicted DO at a certain temperature was above the observed DO at that same temp. These are the places where I _think_ the process guidance could help (as in if it's above 20 deg C, please don't predict DO above ~9 mg/l). 

I see a couple of these in the lower left - there is one prediction of almost 14 mg/l at ~8 deg C; another of ~11 mg/l at ~15 deg C. Those may have have been better with some guidance.

I also see a bit of cluster in the lower center plot at the warmest temps. Maybe those would be improved too.

### The upper sites
The upper two sites with obs are not doing as well. But they are _under_-estimating DO. My take on this possible constraint would be to basically put a "ceiling" on the DO predictions. So in these cases, I don't know that the guidance in terms of the temperature relationship would help improve things.

amcarter:
Are these results from the model trained on DO data or one trained to predict GPP and ER and then calculate DO based on those? In the latter case, the temperature 'constraint' would come in in the calculation step, where we say that DO is the value at saturation (a function of temperature) - ER + GPP and could help to correct for over or underestimation.

>The upper two sites with obs are not doing as well. But they are under-estimating DO. 

We have water temperature for these sites, correct? I know that it is not an input to the model, but is it possible that at these upper sites the water temperature is actually lower than you would predict given the air temperature? That would mean that the model is learning the DO x temp relationship correctly, but making lower estimates than it should because it thinks the water temperature is higher than it is. Could you plot air temp x water temp for these sites and see if the relationship is different in the upper two?

jsadler2:
Thanks for the thoughts, @amcarter 

>Are these results from the model trained on DO data or one trained to predict GPP and ER and then calculate DO based on those?

No. These are not trained to predict GPP and ER.

> Could you plot air temp x water temp for these sites and see if the relationship is different in the upper two?

Great idea. See plots below. (Sorry about the change in site order. I was doing this with UV data before an switched to DV which added a site - also added above).

[...]

Issue #45, opened 2022-01-26
Inspect hidden states for process understanding


jsadler2:
An idea that has come up is to inspect the hidden states to see if they are behaving as we would expect some state or flux in the process would behave. 

The two examples that have come up are:
1. a **biomass**: some representation of a biomass that accumulates in the summer, is lower in the winter, and maybe has a sharp decrease after a scouring event
2. **discharge**: we aren't giving the baseline model discharge as an input. Does the model have a discharge-like hidden state?

We can look into answering these questions with the baseline LSTM model (#40).

jsadler2:
I started looking at this a little bit and it's interesting! A little mysterious, but interesting.

Here are a few plots so far:
## Randomized weights
I first ran the model to get the "DO" predictions and the hidden states with randomized weights to get something to compare them with:

### "DO" preds
![do_rand](https://user-images.githubusercontent.com/6943441/153678772-f34ff458-f696-4b69-8b24-11dc6fa5db57.jpg)

### hidden states
![h_rand](https://user-images.githubusercontent.com/6943441/153678764-ec68d262-cf16-4264-951b-b049ebe757a7.jpg)



## Trained weights
### DO preds
![do](https://user-images.githubusercontent.com/6943441/153678752-a42e97ce-ab1c-47a5-8b93-e6fa1888e5eb.jpg)


### hidden states

![h](https://user-images.githubusercontent.com/6943441/153678735-92e1a833-f78d-4983-8c10-c62a95c80259.jpg)

lekoenig:
Awesome, @jsadler2 - thanks for sharing these! 🚀 

I'm wondering about how/whether I should compare the h time series in the randomized weights plots above versus the trained weights plots below. The weights more or less determine how the inputs are mapped to outputs, right? So in the model run with randomized weights, it seems like the model is learning the relationship between air temperature and DO (e.g. I'm looking at h_2), but that dynamic isn't really maintained in h_2 on the trained weights. It's also interesting that predicted daily mean DO is often greater than the predicted daily max DO for this model run (and that min DO doesn't really have much seasonality) - does that just point to the importance of model training?

For the model run with trained weights, h_3 and h_9 jump out to me as the model appears to be inferring some dynamic that is relatively higher from Oct - May compared to the rest of the year. h_1 looks sort of like a hydrograph (see below), and I can imagine h_0 as some concatenation of the h_3 and h_1. Mysterious but interesting, indeed!

[...]

PR #53, opened 2022-02-23
analysis of baseline states

jsadler2:
This PR is code used to look into:
* #45 
* and #48 

Specifically the code
* outputs the model hidden states to csv files 
* outputs the hidden states across 6 replicates
* makes plots to visualize the hidden state data
* pulls auxiliary data from nwis (flow, water temp, specific conductivity) for comparison with hidden states
* makes a plot that compares predicted and observed DO with water temperature

### DIFF
--- a/3_visualize/src/plot_output_weights.py
+++ b/3_visualize/src/plot_output_weights.py
@@ -1,0 +1,57 @@
+# ---
+# jupyter:
+#   jupytext:
+#     formats: ipynb,py:percent
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.13.7
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %%
+import sys
+import numpy as np
+import matplotlib.pyplot as plt
+
+# %%
+sys.path.insert(0, "../../2a_model/src/models/0_baseline_LSTM/")
+
+# %%
+from model import LSTMModel
+
+# %%
+m = LSTMModel(10, 3)
+
+# %%
+m.load_weights("../../2a_model/out/models/0_baseline_LSTM/train_weights/")
+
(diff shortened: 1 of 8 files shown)


3_visualize/src/plot_output_weights.py
```diff
@@ -0,0 +1,57 @@
...
+sys.path.insert(0, "../../2a_model/src/models/0_baseline_LSTM/")
+
+# %%
+from model import LSTMModel
+
+# %%
+m = LSTMModel(10, 3)
+
+# %%
+m.load_weights("../../2a_model/out/models/0_baseline_LSTM/train_weights/")
```
galengorski, line 31: I'm a little confused where these weights come from

3_visualize/src/plot_output_weights.py
```diff
@@ -0,0 +1,57 @@
...
+
+# %%
+data = np.load("../../2a_model/out/models/0_baseline_LSTM/prepped.npz", allow_pickle=True)
```
galengorski, line 34: Also confused here as to what you are loading, are these input data to use for plotting

galengorski (approved):
This all looks good to me, I just had a few clarifying questions to make sure I understood it. I think I'm still trying to get my head around the code architecture a little bit

3_visualize/src/plot_output_weights.py
jsadler2: These are weights produced by the snakemake workflow:
https://github.com/USGS-R/drb-do-ml/blob/cee3da5b747c035c826641e9bdbb05f91b394a81/2a_model/src/models/0_baseline_LSTM/Snakefile#L71

3_visualize/src/plot_output_weights.py
jsadler2: Yes these are input data. They are used to just run the model once so I can get the weight values. ... now that I think of it, I'm not 100% sure this step is necessary.

Issue #33, opened 2021-12-20
Create data summary plots in R 


lekoenig:
I'm actually kind of thinking we maybe should take out the Python plots too 😄. They kind of take a while to build and then without them we can get rid of those csv targets in 1_fetch.R. They were nice to have to get a feel for the data, but I'm not sure we need to keep them around. I could be pursuaded otherwise tho about both the python targets and the plots from r.

Maybe we do a happy medium and produce one summary plot with ggplot (to avoid the python rigmarole)

_Originally posted by @jsadler2 in https://github.com/USGS-R/drb-do-ml/issues/32#issuecomment-998305511_

lekoenig:
- [ ] Create summary plots in R/ggplot2
- [ ] Get rid of `p1_daily_data_csv` and `p1_inst_data_csv` targets
- [ ] Get rid of python code to create plots: `3_visualize/src/do_overview_plots.py`

PR #58, opened 2022-03-04
33 data summary plots

lekoenig:
Addresses #33.

Namely, code changes here move data summary plots from python to R so that we can simplify the pipeline a bit; we get rid of two csv targets in the fetch phase and remove one place where we need `reticulate`.


jsadler2 (approved):
Looks great. Thanks!

Issue #50, opened 2022-02-14
why is 01481500 "tidal"?


jsadler2:
@jsadler2, do you remember how you determined `01481500` is tidal?

_Originally posted by @lekoenig in https://github.com/USGS-R/drb-do-ml/pull/40#discussion_r801115087_

jsadler2:
Huh. I'm glad you brought this up, @lekoenig. When I look at the specific conductance data I'm not seeing a tidal signal after all. I'm not sure why I thought that it was tidal originally, but looking at this, it seems like it's not.
![image](https://user-images.githubusercontent.com/6943441/155412273-ebc2026d-e393-4e28-b17c-32852e27a0f7.png)

lekoenig:
Thanks for taking another look at this site. So are we in agreement to add "01481500" back in the mix?

jsadler2:
I agree

PR #60, opened 2022-03-11
Retain nwis 01481500

lekoenig:
Addresses #50.

In #50, we could not identify a reason why [01481500](https://waterdata.usgs.gov/usa/nwis/uv?01481500) was characterized as tidal, so I removed that site from `omit_nwis_sites`. This change retains 01481500 in our data pipeline, and it is now included in our list of well-observed sites (now 14 sites; number of obs-days > 300). Note that I have not specified 01481500 within any of our train/test/val splits in this PR.


jsadler2 (approved):
Welcome back, 01481500

Issue #66, opened 2022-03-14
Error when building p3_well_observed_site_data_json


lekoenig:
When I try to run `tar_make` to build the pipeline I'm getting an error associated with `p3_well_observed_site_data_json`:

```
* start target p3_well_observed_site_data_json
GeoJSON' layer not supported by driver `
Deleting layer `well_observed_trn_val_test' failed
Writing layer `well_observed_trn_val_test' to data source 
  `3_visualize/out/well_observed_trn_val_test.geojson' using driver `GeoJSON'
Updating existing layer well_observed_trn_val_test
Writing 14 features with 12 fields and geometry type Point.
Unknown field name `bird_dist_to_subseg_m': updating a layer with improper field name(s)?
x error target p3_well_observed_site_data_json
```

jsadler2:
good news ... kind of ... i'm now getting the same error

lekoenig:
@jsadler2, can we close this issue?

PR #72, opened 2022-03-16
[#66] delete_layer in geojson

jsadler2:
seems like this fixes the error in writing the geojson

### DIFF
--- a/3_visualize.R
+++ b/3_visualize.R
@@ -65,7 +65,7 @@ p3_targets_list <- list(
     p3_well_observed_site_data_json,
     {
       filename = "3_visualize/out/well_observed_trn_val_test.geojson"
-      st_write(p3_well_observed_site_data, filename, append=FALSE)
+      st_write(p3_well_observed_site_data, filename, append=FALSE, delete_layer=TRUE)
       filename
     },
     format = "file"
--- a/3_visualize.R
+++ b/3_visualize.R
@@ -65,7 +65,7 @@ p3_targets_list <- list(
     p3_well_observed_site_data_json,
     {
       filename = "3_visualize/out/well_observed_trn_val_test.geojson"
-      st_write(p3_well_observed_site_data, filename, append=FALSE, delete_layer=TRUE)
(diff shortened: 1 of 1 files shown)


jsadler2:
@lekoenig - can you give this a try?

lekoenig:
Yep, thanks for looking into this! Glancing at #66 again, I must have been trying to build on my branch `36-match-segs` because one of those errors is related to the variable name change I made in PR #67 that's not yet in main (i.e. `offset` --> `bird_dist_to_subseg_m`).

3_visualize.R
```diff
@@ -65,7 +65,7 @@ p3_targets_list <- list(
     p3_well_observed_site_data_json,
     {
       filename = "3_visualize/out/well_observed_trn_val_test.geojson"
-      st_write(p3_well_observed_site_data, filename, append=FALSE)
+      st_write(p3_well_observed_site_data, filename, append=FALSE, delete_layer=TRUE)
```
lekoenig, line 68: The geojson file may be writing OK, but `delete_layer = TRUE` still leaves me with these errors:
```
GeoJSON' layer not supported by driver `
Deleting layer `well_observed_trn_val_test' failed
```

After checking out the [documentation](https://r-spatial.github.io/sf/reference/st_write.html) for `st_write`, I wonder if instead we need to specify `delete_dsn`. I'm curious if the change below works for you (I also made the driver explicit and added a quiet message to suppress verbose output when running `tar_make`:

```suggestion
      st_write(p3_well_observed_site_data, filename, append = FALSE, delete_dsn = TRUE, driver = "GeoJSON", quiet = TRUE)
```

lekoenig (commented):
I'm still getting the geojson error with `delete_layer = TRUE` but let me know what you think about the suggestion I've added here.

3_visualize.R
jsadler2: That works great! Thanks for looking into this, @lekoenig

lekoenig (approved):
Looks good! Thanks for checking on this.

Issue #73, opened 2022-03-16
Create target for "medium_observed" sites


lekoenig:
Following up on #49, we'll need a target that reflects "medium-observed" sites (i.e., sites with at least 100 observation-days) to assess the impact of lowering our threshold for including a site in the LSTM model.

lekoenig:
Do we want to stick with just using NWIS sites? There are 23 NWIS sites that have at least 100 obs-days worth of data. Looking across our entire site list, however, there are 113 discrete + NWIS sites with at least 100 obs-days worth of data because some discrete sites are sampled at ~regular intervals, e.g.:

![DO_example](https://user-images.githubusercontent.com/8785034/159699408-df7d4c01-fb25-4afc-86b8-84ec14f498aa.png)

PR #78, opened 2022-03-23
add new targets for moderately-observed sites

lekoenig:
Addresses #73.

The code changes here add a new target, `p2_med_observed_sites` that indicates the NWIS sites with at least 100 observation-days. Up to this point we have only considered well-observed sites (>300 obs-days; n = 14). By lowering our criteria for the number of observations, we get 23 "moderately-observed" sites:

```
> p2_well_observed_sites
 [1] "01466500" "01472104" "01472119" "01473499" "01473500" "01473675" "01473780" "01474500" "01475530" "01475548" "01480617" "01480870" "01481000"
[14] "01481500"
> 
> p2_med_observed_sites
 [1] "01465798"  "01466500"  "01467042"  "01467048"  "01467086"  "01467087"  "01472104"  "01472119"  "014721254" "014721259" "01473499"  "01473500" 
[13] "01473675"  "01473780"  "01473900"  "01474000"  "01474500"  "01475530"  "01475548"  "01480617"  "01480870"  "01481000"  "01481500" 
> 
```

This PR also adds a new "medobs" version of all the targets in `2a_model.R` (using the same `val_sites` and `tst_sites` defined in `_targets.R` as before), so new zarr files get written to reflect these moderately-observed sites:

network_diagram

A couple questions:

- I've assumed we want to stick with the NWIS sites, but there are discrete sites that meet this new criteria (see #73). Do we want to include those in the training data?
- Incorporating the moderately-observed sites makes the `2a_model.R` file pretty bulky since we're essentially repeating targets. I would welcome any suggestions you have for how to clean this up, e.g. perhaps we could add a toggle that indicates whether we want to write the zarr files for well-observed vs. moderately-observed. We could also keep as is.

### DIFF
--- a/2a_model.R
+++ b/2a_model.R
@@ -137,6 +137,84 @@ p2a_targets_list <- list(
     },
     format="file",
     pattern = map(p2a_model_ids)
+  ),
+  
+  
+  ## CREATE EQUIVALENT TARGETS FOR "MODERATELY-OBSERVED SITES" ##
+  # char vector of moderately-observed train sites
+  tar_target(
+    p2a_trn_sites_medobs,
+    p2_med_observed_sites[!(p2_med_observed_sites %in% val_sites) & !(p2_med_observed_sites %in% tst_sites)]
+  ),
+  
+  # char vector of moderately-observed val and training sites
+  tar_target(
+    p2a_trn_val_sites_medobs,
+    p2_med_observed_sites[(p2_med_observed_sites %in% p2a_trn_sites_medobs) | (p2_med_observed_sites %in% val_sites)]
+  ),
+  
+  # get moderately-observed sites that we use for trning, but also have data in the val time period
+  tar_target(
+    p2a_trn_sites_w_val_data_medobs,
+    p2_daily_with_seg_ids  %>%
+      filter(site_id %in% p2a_trn_val_sites_medobs,
+             !site_id %in% val_sites,
+             date >= val_start_date,
+             date < val_end_date) %>%
+      group_by(site_id) %>%
+      summarise(val_count = sum(!is.na(do_mean))) %>%
+      filter(val_count > 0) %>%
+      pull(site_id)
+  ),
+  
+  # moderately-observed sites that are trning sites but do not have data in val period
+  tar_target(
+    p2a_trn_only_medobs,
+    p2a_trn_sites_medobs[!p2a_trn_sites_medobs %in% p2a_trn_sites_w_val_data_medobs]
+  ),
+  
+  
+  ## WRITE OUT PARTITION INPUT AND OUTPUT DATA ##
+  # write trn met and seg attribute data to zarr
+  # note - I have to subset before passing to subset_and_write_zarr or else I
+  # get a memory error on the join
+  tar_target(
+    p2a_trn_inputs_medobs_zarr,
+    { 
+      trn_input <- p2a_met_data_w_sites %>%
+        filter(site_id %in% p2a_trn_sites_medobs) %>%
+        inner_join(p2a_seg_attr_w_sites, by = "site_id")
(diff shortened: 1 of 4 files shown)


2a_model.R
```diff
@@ -101,6 +101,84 @@ p2a_targets_list <- list(
...
+      summarise(val_count = sum(!is.na(do_mean))) %>%
+      filter(val_count > 0) %>%
+      pull(site_id)
+  ),
+  
+  # moderately-observed sites that are trning sites but do not have data in val period
+  tar_target(
+    p2a_trn_only_medobs,
+    p2a_trn_sites_medobs[!p2a_trn_sites_medobs %in% p2a_trn_sites_w_val_data_medobs]
+  ),
```
jsadler2, line 138: These two targets are only used in making a sf object in 3_visualize:
https://github.com/USGS-R/drb-do-ml/blob/6e0afdcc0aed84b7aae1d3497775ff5835e12faf/3_visualize.R#L51-L62

So I'm not entirely sure we should have them (or the ones for the "well_obs" sites) in 2a. It may actually be better just to make them as part of the `p3_well_observed_site_data` target if we end up making that. But for now, I'd suggest leaving these out.

2a_model.R
```diff
@@ -101,6 +101,84 @@ p2a_targets_list <- list(
...
+    p2a_trn_do_medobs_zarr,
+    subset_and_write_zarr(p2_daily_with_seg_ids, "2a_model/out/med_observed_trn_do.zarr", p2a_trn_sites_medobs),
+    format="file"
+  ),
+  
+  # write trn and val do data to zarr
+  tar_target(
+    p2a_trn_val_do_medobs_zarr,
+    subset_and_write_zarr(p2_daily_with_seg_ids, "2a_model/out/med_observed_trn_do.zarr", p2a_trn_val_sites_medobs),
+    format="file"
```
jsadler2, line 181: We'll only need _one_ target now: `p2a_med_obs_data` - like this target but with the med_obs_sites:
https://github.com/USGS-R/drb-do-ml/blob/6e0afdcc0aed84b7aae1d3497775ff5835e12faf/2a_model.R#L73

jsadler2 (changes requested):
A couple comments below. 

We should also think about adding 2-3 sites from these new ones to the set of validation sites. I don't think we need to be too particular about these, but having some variety in geography would be good.

Also, did you screen these to make sure we aren't getting tidally-influenced sites?

2a_model.R
lekoenig: OK, good to know. I omitted the equivalent of these targets since it makes the 2a_model phase file pretty bulky. Instead, I just filtered the met data for sites that are within `p2_med_observed_sites` and _not_ within the `tst_sites` vector defined in `_targets.R`.

lekoenig:
Also, I made a small change in https://github.com/USGS-R/drb-do-ml/pull/78/commits/86b0cfe409fda5d4a13deb57226a6a3e8ce97d7e to update the dplyr syntax in `get_site_flowlines` (one of the warnings noted in #83). I checked that the results of that function did not change as a result of this formatting update.

jsadler2 (approved):
~~Looks good. Thanks, @lekoenig !~~ (see one request below)

_targets.R
```diff
@@ -62,7 +62,7 @@ earliest_date <- "1979-10-01"
 dummy_date <- "2021-12-19"
 
 # test and validation sites
-val_sites <- c("01472104", "01473500", "01481500")
+val_sites <- c("01472104", "01473500", "01481500","014721254")
```
jsadler2, line 65:
```suggestion
val_sites <- c("01472104", "01473500", "01481500")
```

jsadler2 (changes requested):
Actually. With just two sites added. I don't think we should add either to the validation sites.

Issue #79, opened 2022-03-31
Create function to estimate daily (normalized) light


lekoenig:
For the process guidance equations for DO-max outlined in #47, we want some estimate of daily light to include in the model _inputs._ Two options we've discussed:

- write a wrapper function that uses `calc_light()` in `streamMetabolizer` to estimate a single value of Lt/L-sum for each day. This value represents the daily maximum light relative to the daily cumulative light.
- either estimate or grab the estimates of `day.length` from the Appling data release and use the ratio of day length / 24 hours as a proxy for the ratio of max light to the daily sum.

PR #81, opened 2022-04-01
Add function to calculate daily normalized light values

lekoenig:
Addresses #79.

The function `calc_daily_light()` calculates daily normalized light values given a range of dates and a lat/lon location. The output includes two proxies for relative light that we discussed on 3/31/2022:

- The daily maximum "instantaneous" light relative to the cumulative daily light (L_t / sigma-L),  as in #47
- a metric that relies on day length; I've been thinking of this as a correction based on day.length/24 hours, but I realized it's maybe more appropriate to calculate delta-t/24 (here, we assume delta-t equals 30 min) to represent the fraction of day length received during the 30-min window that coincides with the time of DO-max. 

@amcarter, I'd be curious if this is similar to the approach you were taking to calculating light, or if you have any other thoughts about the two proxies listed above. Both approaches are returned in the target `p2_daily_max_light` and are indicated by `frac_light` and `frac_daylength`, respectively:

```
head(p2_daily_max_light)
# A tibble: 6 x 8
  date_localtime max_light sum_light day_length frac_light frac_daylength subsegid seg_id_nat
  <date>             <dbl>     <dbl>      <dbl>      <dbl>          <dbl> <chr>         <int>
1 1979-09-30            0         0        NA     NaN             NA      1_1            1435
2 1979-10-01         1621.    24008.       11.5     0.0675         0.0435 1_1            1435
3 1979-10-02         1609.    23736.       11       0.0678         0.0455 1_1            1435
4 1979-10-03         1597.    23472.       11       0.0680         0.0455 1_1            1435
5 1979-10-04         1585.    23219.       10.5     0.0683         0.0476 1_1            1435
6 1979-10-05         1573.    22967.       10.5     0.0685         0.0476 1_1            1435
> 
```

`p2_daily_max_light` maps the `calc_daily_light()` over each of the 459 segments in the network and took quite awhile to build (~1.5 hrs on my local machine). I'm open to any suggestions you have for cutting down on that run time. 

For now, I've gone with the first approach when integrating the "light ratio" with our dynamic met drivers in `2a_model.R`; the code changes here add a new target `p2a_met_light_data` that combines the met data with the light data before matching to seg id's and outputting the input data files. 

Finally, for some reason my target `p1_nwis_sites_inst` became outdated and I encountered a new error that had to be addressed to get the pipeline to build. See #82 for further details.


jsadler2 (approved):
this looks good to me and has been working. Thanks, @lekoenig

PR #89, opened 2022-04-13
77 reproduce preds

jsadler2:
This addition uses targets to do the following for each experiment specified:
1. produce the `prepped.npz` file so that it can be used to make the model predictions.
2. call snakemake to "touch" the "trained_weights" files so that they appear up to date
3. call snakemake to produce the final metrics file and all intermediate files including the predictions

### DIFF
--- a/2a_model/src/models/Snakefile_base.smk
+++ b/2a_model/src/models/Snakefile_base.smk
@@ -22,36 +22,11 @@ rule as_run_config:
         asRunConfig(config,output[0])
 
 
-rule prep_io_data:
-    input:
-        "../../../out/well_observed_trn_inputs.zarr",
-        "../../../out/well_observed_trn_targets.zarr",
-    output:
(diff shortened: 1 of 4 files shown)


2a_model/src/models/Snakefile_base.smk
```diff
@@ -22,36 +26,11 @@ rule as_run_config:
         asRunConfig(config,output[0])
 
 
-rule prep_io_data:
```
galengorski, line 25: So is this moving `prep_io_data` from snakemake to targets, is that right?

galengorski (approved):

2a_model/src/models/Snakefile_base.smk
jsadler2: Right.

galengorski: Awesome, thanks!

jsadler2:
Superseded by #90

PR #90, opened 2022-04-13
56 lstm metab

### LINKED ISSUE #56: implement DO model with explicit metabolism representation
Implement the changes proposed here https://github.com/USGS-R/drb-do-ml/discussions/47#discussioncomment-2083081 to explicitly represent metabolism within the DO model.

image

### LINKED ISSUE #77: reproduce trained model predictions using targets pipeline
we need some code in the targets pipeline to reproduce model predictions.

this code will need to do a few things:
- generate the `prepped.npz` file using the `prep_all_data` function from river_dl
- loop through the directories and find all of the `trained_weights` subdirectories
- use the predict functions from river_dl to make predictions from the model

### LINKED ISSUE #87: try lstm-metab with just DO
For diagnosing the lstm-metab, I think it would be useful to see what happens when we just have the model train to just the DO observations.

### LINKED ISSUE #88: try lstm-metab with just metabolism estimates
For diagnosing the lstm-metab, I think it would be useful to see what happens when we just have the model train to just the metabolism estimates

### LINKED ISSUE #92: metab mutli-task
I think it would be interesting to see what effect solving for all of the outputs at once would have instead of having the model solve for the metabolism first and then using the PB model to get to DO.

### LINKED ISSUE #93: Try LSTM-metab with metab plus DO mean
In #88, the DO mean is being affected by the random starting weights. That doesn't make a lot of sense. It would make more sense to tune that parameter.

### LINKED ISSUE #94: Try metab multi-task with just GPP/ER
@amcarter suggested trying the metabolism multi-task with just DO and GPP and ER.

### LINKED ISSUE #96: Try lstm-metab with a little more flexibility
based on the results from #56, #87, and #93, I would like to try the LSTM-metab with a little more flexibility for how it lands on DO min and DO max

jsadler2:
Pretty massive PR. Most of the 58 files changed are actually just moving the previous train_weights to a sub directory so we can keep track of number of training epochs. I'll annotate some of the changes in the "changes" tab to hopefully help with understanding of the changes.

This PR includes functions in #77 and diverged some from that branch. I will therefore close #89.

closes #77
closes #56 
closes #88 
closes #87
closes #92 
closes #93
closes #96
closes #94


2a_model.R
```diff
@@ -67,54 +67,92 @@ p2a_targets_list <- list(
...
+  tar_target(
+    p2a_prepped,
     {
-      # need to join the metab data with the DO observations. 
-      do_and_metab <- p2_daily_with_seg_ids %>%
-          left_join(p2_metab_filtered, by=c("site_id", "date"))
-      subset_and_write_zarr(do_and_metab, "2a_model/out/well_observed_trn_val_targets.zarr", p2a_trn_val_sites)
+    dir.create(sprintf("2a_model/out/models/%s", p2a_model_ids), showWarnings = FALSE)
+    prep_io_data(x_data_file = p2a_trn_inputs_zarr,
+                 y_data_file = p2a_trn_targets_zarr,
```
lekoenig, line 111: should x_data_file and y_data_file here reference `p2a_well_obs_inputs_zarr`? I may be confusing some of the target names, but wasn't `p2a_trn_inputs_zarr` replaced with `p2a_well_obs_inputs_zarr` above?

lekoenig, line 111: should x_data_file and y_data_file here reference `p2a_well_obs_inputs_zarr`? I may be confusing some of the target names, but wasn't `p2a_trn_inputs_zarr` replaced with `p2a_well_obs_inputs_zarr` above?

jsadler2: Yes. I think you are right.

lekoenig, line 111:
```suggestion
    prep_io_data(x_data_file = p2a_well_obs_inputs_zarr,
                 y_data_file = p2a_well_obs_targets_zarr,
```

PR #108, opened 2022-05-25
107 - cleanup

### LINKED ISSUE #97: Use same val time periods for baseline and lstm-metab models
The lstm-metab models have a smaller validation time period b/c there isn't as much metabolism data. We should consider revisiting the baseline model and using the same validation period throughout.

### LINKED ISSUE #103: bug 1 in using metabolism data: left_join
Comparing these two plots made me realize there are two bugs in our code for bringing in the metabolism data.
(Lauren created this one from `p1_metab`)
image
(I created this one from `2a_model/out/well_obs_targets.zarr`)
image

### bug 1. `left_join` should be `full_join`. 
In the following lines we do a `left_join`: 
https://github.com/USGS-R/drb-do-ml/blob/c495411c579dd76c78dabc0c346893c2e17b249f/2a_model.R#L90-L93

That means we are only including metabolism estimates where we have do data. That is why I have a blank for the `01474500` and Lauren doesn't. Mine is post-join, and there aren't any metabolism estimates that line up with DO obs in that site. So we should do a `full_join` instead so we don't lose any data.
 
See next issue for bug 2

### LINKED ISSUE #104: bug 2 in using metabolism data: not using filtered data
continuing from #103 

In the following lines, we are joining `p1_metab` to the other meteorological variables. 

https://github.com/USGS-R/drb-do-ml/blob/c495411c579dd76c78dabc0c346893c2e17b249f/2a_model.R#L90-L93

However, we _should_ be joining `p2_metab_filtered`. 

We also will need to either change the name of the columns from `GPP_filtered` to `GPP` and `ER_filtered` to `ER` OR we need to change the names in the config file:

https://github.com/USGS-R/drb-do-ml/blob/c495411c579dd76c78dabc0c346893c2e17b249f/2a_model/src/models/2_metab_multitask/config.yml#L12

### LINKED ISSUE #107: cleaning up repo
I would like to do some general clean up in the repo. A lot of what we have tried so far has been very experimental and diverging which I think has been good. But given the short fuse at this point, I think it would be good to start converging. 

I would like to address
* #97 
* #103 
* #104 
* #106 - maybe

Because we are producing new model results with these changes, I suggest that we just let the old model results live in the git history and overwrite with the new. I also suggest changing the number slightly. As part of this, I propose we just leave what was v1 (LSTM-metab) in the git history since it never worked and I don't think we have time to try that any more for this paper.

old model version | new model version
-- | --
v0 | v0 (no change)
v1 | - 
v2 | v1
v3 | -

I would also like to have several replicates (maybe 10) for each model run.

Finally, I think it makes sense to have a base config file and then that can be augmented by version-specific config files

jsadler2:
Cleaning up several things, the big one is that data from validation sites was being used in the multitask models. 

See #107 for more details on what is being cleaned up here.

Closes #107 
Closes #104 
Closes #103 
Closes #97


2a_model/src/models/0_baseline_LSTM/config.yml
```diff
@@ -1,45 +1,8 @@
-out_dir: "../../../out/models"
 exp_name: "0_baseline_LSTM"
 
-#### for flow modeling, seg_width should be removed from x_vars
```
jsadler2, line 4: I'm leaving only the experiment-specific options in this file (e.g., "exp_name", "y_vars") because its the "experiment-specific" config file that extends the "base" config file.

2a_model/src/models/Snakefile_base.smk
```diff
@@ -27,6 +27,46 @@ rule as_run_config:
         asRunConfig(config,output[0])
 
 
+rule prep_io_data:
```
jsadler2, line 30: I put this back in the Snakefile. I had originally thought of not mixing Snakemake and Targets which is why I had this in Targets before (so we didn't have to call Snakemake to produce it). Since then, I realized it was going to be too tricky to try to circumvent Snakemake altogether. Since we are calling Snakemake anyway here (https://github.com/USGS-R/drb-do-ml/blob/c495411c579dd76c78dabc0c346893c2e17b249f/2a_model.R#L152), I figured it made sense to have this python function call back in snakemake.

2a_model/src/models/Snakefile_base.smk
```diff
@@ -27,6 +27,46 @@ rule as_run_config:
...
+                      y_vars_finetune=config['y_vars'],
+                      spatial_idx_name='site_id',
+                      time_idx_name='date',
+                      train_start_date=config['train_start_date'],
+                      train_end_date=config['train_end_date'],
+                      val_start_date=config['val_start_date'],
+                      val_end_date=config['val_end_date'],
+                      test_start_date=config['test_start_date'],
+                      test_end_date=config['test_end_date'],
+                      val_sites=config['validation_sites'],
```
jsadler2, line 49: This line was not here before. So even though we specified the "validation_sites" in the configuration file, those sites were not being passed to the `prep_all_data` function which means they were not removed from the training data.

2a_model/src/models/Snakefile_base.smk
```diff
@@ -27,6 +27,46 @@ rule as_run_config:
...
+        data = np.load(output[0], allow_pickle=True)
+        df_trn = prepped_array_to_df(data['y_obs_trn'],
+                                     data['times_trn'],
+                                     data['ids_trn'],
+                                     col_names=data['y_obs_vars'])
+        df_trn_val_sites = df_trn[df_trn.seg_id_nat.isin(config['validation_sites'])]
+        df_trn_tst_sites = df_trn[df_trn.seg_id_nat.isin(config['test_sites'])]
+
+        assert df_trn_val_sites['do_mean'].notna().sum() == 0
+        assert df_trn_tst_sites['do_mean'].notna().sum() == 0
```
jsadler2, line 66: These lines are here now to make sure that there is no data in the training dataset for the validation sites. If there is data in any of the sites, it will throw an error. I've tested this and it works.

jsadler2:
@galengorski - can you look this PR over. Most of the 198 files thankfully do not need review.

In the File Filter in the Files Tab you can filter out all of the "dotfiles" and the "deleted files." There is one Snakefile under the "no extension" files. That should leave 11 files to review and most of the changes are pretty minor (though the validation sites is obviously  very consequential). I've left some comments at different points in the files to explain some of the changes.

image

galengorski (approved):

Issue #106, opened 2022-05-20
really off predictions at 01480617


jsadler2:
there are some really strange predictions at 01480617 for model version 1a
![image](https://user-images.githubusercontent.com/6943441/169602552-92f22b54-a38c-4c2c-985e-3c6fdb6fcf65.png)
![image](https://user-images.githubusercontent.com/6943441/169602623-a3bb66de-f320-4f8c-bf33-a037b2f708e3.png)

I'm not sure how/why this is happening ... especially since the predictions are being made through the process-based equations ...

jsadler2:
I dug into this some more and I think I found the culprit: very small depth estimates.


![image](https://user-images.githubusercontent.com/6943441/169616231-ddc0bd5b-78c2-4b07-abd7-1a6393b22e39.png)

The depth estimates are on the bottom of the equations:
![image](https://user-images.githubusercontent.com/6943441/169615239-50346db1-9337-4e19-b702-4d5b6cc3cdfc.png)

So when those get really small, the DO min/max values get really big. The max value is even bigger b/c delta t is also in the denominator and that is 1/48.

I'm not sure if there is a good way to fix this. Maybe we make sure the depth value is at least some minimum number (e.g., 0.1 m)? Or just add some small constant to the denominator?

PR #110, opened 2022-05-27
[#106] filter daily light to well-observed reaches

### LINKED ISSUE #109: Filter reaches when building p2_daily_max_light
We currently build `p2_daily_max_light` by mapping over all 459 DRB reaches in `p1_reaches_sf`. This means it takes a long time (>1 hr) to build this target. To simplify the target and speed up the build time we should filter `p1_reaches_sf` to just the reaches with observations.

jsadler2:
Make light estimates for just segments with well-observed sites. 

I don't love this implementation b/c it doesn't lend itself well to expanding to other sites, but I don't see an easy way to make that more flexible.

closes #109

### DIFF
--- a/2_process.R
+++ b/2_process.R
@@ -9,6 +9,8 @@ source("2_process/src/calc_daily_light.R")
 source("2_process/src/metab_utils.R")
 source("1_fetch/src/write_data.R")
 
+library(sf)
+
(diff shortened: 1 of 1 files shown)


jsadler2:
Closing since I was referring to the wrong issue (106 instead of 109)

PR #111, opened 2022-05-27
combine snakemake rules; model_ids list of lists

jsadler2:
This combines the prep_data creation, the "touch" step, and the actual model run step. 

Also, with the inclusion of the base_config file (see #108), the p2a_model_ids needed to be adjusted. 

So now it is a list of lists where we can specify the model_id, the snakefile directory, and the path to the config file.


jsadler2:
merge #112 first

PR #112, opened 2022-05-27
[#109] filter daily light to well-observed reaches

jsadler2:
Make light estimates for just segments with well-observed sites.

I don't love this implementation b/c it doesn't lend itself well to expanding to other sites, but I don't see an easy way to make that more flexible.

closes https://github.com/USGS-R/drb-do-ml/issues/109

### DIFF
--- a/2_process.R
+++ b/2_process.R
@@ -9,6 +9,7 @@ source("2_process/src/calc_daily_light.R")
 source("2_process/src/metab_utils.R")
 source("1_fetch/src/write_data.R")
 
+# Explicitly attach sf package to handle geometry data when mapping over `p1_reaches_sf`
 library(sf)
 
 p2_targets_list <- list(
@@ -96,7 +97,7 @@ p2_targets_list <- list(
  
  # filter p1_reaches_sf to segments with "well-observed" sites
  tar_target(   
-   p2_well_obs_reaches,
+   p2_well_observed_reaches,
    {
    well_obs_reach_ids <- p2_sites_w_segs %>%
      filter(site_id %in% p2_well_observed_sites) %>% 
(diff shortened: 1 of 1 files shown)


2_process.R
```diff
@@ -9,6 +9,8 @@ source("2_process/src/calc_daily_light.R")
 source("2_process/src/metab_utils.R")
 source("1_fetch/src/write_data.R")
 
+library(sf)
```
lekoenig, line 12:
```suggestion
# Explicitly attach sf package to handle geometry data when mapping over `p1_reaches_sf`
library(sf)
```

lekoenig: ^ A small suggestion to remind ourselves why the call to `library(sf)` is needed here. Also, do you think it makes sense to move this line to `_targets.R`? I think I'm more likely to see it there, and that would keep our config-type stuff in one place.

2_process.R
```diff
@@ -85,20 +87,33 @@ p2_targets_list <- list(
...
 
   # make list of "well-observed" sites
   tar_target(
    p2_well_observed_sites,
    p2_sites_w_segs %>% filter(count_days_total > 300) %>% pull(site_id)
  ),
  
+ # filter p1_reaches_sf to segments with "well-observed" sites
+ tar_target(   
+   p2_well_obs_reaches,
```
lekoenig, line 99:
```suggestion
   p2_well_observed_reaches,
```

Naming suggestion to be consistent with `p2_well_observed_sites` above.

lekoenig: If you opt to make this change, `p2_well_obs_reaches` should also be edited below (lines 112 and 114).

lekoenig (approved):
This builds for me and is now much faster! 🎉 

```
> tar_meta(name = p2_daily_max_light)$seconds
[1] 134.71
```

2_process.R
```diff
@@ -85,20 +87,33 @@ p2_targets_list <- list(
...
+   }
+ ),
+  
+  # Estimate daily (normalized) max-light
+  tar_target(
+    p2_daily_max_light,
+    { 
+    calc_seg_light_ratio(p2_well_obs_reaches, start_date = earliest_date, end_date = dummy_date)
+    },
+    pattern = map(p2_well_obs_reaches)
```
jsadler2, line 114:
```suggestion
    calc_seg_light_ratio(p2_well_observed_reaches, start_date = earliest_date, end_date = dummy_date)
    },
    pattern = map(p2_well_observed_reaches)
```

PR #114, opened 2022-06-03
113 multitask dense

### LINKED ISSUE #113: try multitask model with a dense layer between metab and DO
I think it would be interesting to force the model to use the metabolism predictions to predict DO. In my mind this is one step above the multitask model in terms of process guidance. In the multitask model, its true that the same hidden states have to be used to make predictions of DO and metabolism, but there's no explicit link between the variables. If we add a dense layer that takes metabolism as input and give DO as output, it means that the DO predictions are explicitly _dependent_ on the metabolism estimates (though the relationship between the two is still up to the model).

jsadler2:
multitask dense model

closes #113


galengorski (approved):
Looks good, this is the one we all did together, so I'm on board

PR #124, opened 2022-06-14
Add NHDv2 flowlines and match sites to segments

### LINKED ISSUE #120: Load NHDv2 flowlines
Add a target that downloads the NHDv2 flowlines for the Delaware River Basin

### LINKED ISSUE #121: Match DO sites to NHDv2 flowlines
Modify the `get_site_flowlines` function (in 2_process/src/match_sites_reaches) to match DO sites to NHDv2 flowlines.

lekoenig:
This PR adds NHDv2 flowlines to the pipeline (`p1_nhd_reaches_sf`) and matches the DO sites to the downloaded NHD segments (`p2_sites_w_nhd_segs`). 

A new function, `get_site_nhd_flowlines()`, was added to match the DO sites to the flowlines. This function should look familiar, it's a slightly simpler version of the previous `get_site_flowlines()`, which required a couple extra steps to modify the nhdplusTools helper functions for use with the NHM network. 

I've also done some ~light QC on the site to nhd seg matches themselves. For instance, for one of our "well-observed" sites on the Schuylkill, the matching function picks up a nearby tributary rather than the mainstem river: 

schuylkill

I compared our matched COMIDs with those in [ref-gages](https://github.com/internetofwater/ref_gages) v0.6, subset the (USGS) sites that had different COMIDs estimated by `get_site_nhd_flowlines` vs ref-gages, and visually inspected those. Ultimately, my goal is to bring ref-gages into the pipeline to QC and reconcile site to reach matches. However, this dataset is fairly new and there are many sites where the ref-gages estimate isn't right either. Given our timeline for the June '22 mini-sprint, I opted to instead add a file `1_fetch/in/refgages_manual` that represents our source of truth that should be used over the output from `get_site_nhd_flowlines` for 30 sites (including 2 well-obs sites and 4 med-obs sites). 

I haven't yet replaced `p2_sites_w_segs` with `p2_sites_w_nhd_segs` in downstream code, and can either do that as part of this PR or in a separate one. 

Closes #120
Closes #121


2_process/src/match_sites_reaches.R
```diff
@@ -78,3 +80,71 @@ get_site_flowlines <- function(reach_sf, sites, sites_crs, max_matches = 1, sear
...
+#' @param max_matches the maximum number of segments that a point can match to
+#' @param search_radius the maximum radius in meters to match a point to a segment;
+#' segments outside of search_radius will not match
+#'
+#' @value A data frame that contains the same columns as the `sites` data,
+#' with additional columns `COMID`, `bird_dist_to_comid_m` where 
+#' `bird_dist_to_comid_m` is the distance (in meters) between the site and the
+#' matched NHDv2 flowline.
+#'
+get_site_nhd_flowlines <- function(nhd_lines, sites, sites_crs,
```
jsadler2, line 99: Why did you decide to make a new function, `get_site_nhd_flowlines` and not use `get_site_flowlines`? The two look very similar.

jsadler2 (changes requested):
Thanks, Lauren ... especially for the manual QC. I do have one question that I think we should think about before merging.

jsadler2:
PS. I think we should have a separate PR that takes out the targets related to PRMS segments.

2_process/src/match_sites_reaches.R
lekoenig: I made a new function because originally I wasn't going to delete the equivalent PRMS/NHM target, `p2_sites_w_segs` (at least not in this PR). These two functions are nearly identical aside from formatting that was needed to adapt the nhdplusTools function to the NHM fabric. We can delete `get_site_flowlines` if we're opting to remove all the NHM stuff from the pipeline. Should I go ahead and do that in this PR?

PR #126, opened 2022-06-14
Delete NHM network adjacency matrix targets

### LINKED ISSUE #119: Omit NHM network adjacency matrix targets
We don't use two targets related to the NHM network adjacency matrix (`p1_ntw_adj_matrix_csv` and `p1_ntw_adj_matrix`) anywhere in the pipeline and we're not likely to do so if we migrate to NHDv2. To simplify the 1_fetch phase file, I suggest we omit these targets.

lekoenig:
This PR deletes two targets related to the network adjacency matrix for the NHM fabric. We weren't using them anyway, and we don't expect to use them when we move to the NHD fabric, either. 

Closes #119

### DIFF
--- a/1_fetch.R
+++ b/1_fetch.R
@@ -182,21 +182,6 @@ p1_targets_list <- list(
     p1_seg_attr_data,
     arrow::read_feather("1_fetch/in/seg_attr_drb.feather")
   ),
-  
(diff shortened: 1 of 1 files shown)


jsadler2 (approved):
+1

Issue #131, opened 2022-06-15
s3 target for drb gridmet


jsadler2:
make an S3 target for the drb gridmet data.

lekoenig:
@jsadler2 tried this in #133 and #137. 

{targets} x S3 isn't working as we expected for file targets like the .nc file containing the gridmet met drivers aggregated for all the NHDv2 catchments in the DRB.  For example, when running `tar_make()`, I would have expected targets to go and download the file from S3 if it is not stored locally (and place it in the given file path).

```
# target definition:  
tar_target(
    p1_drb_nhd_gridmet,
    "1_fetch/in/drb_climate_2022_06_14.nc",
    format = "file"
  )
```

It doesn't do that, but rather, throws a complaint that the file isn't available in the local path. 

```
# run tar_make() to download the file from S3
> tar_make(p1_drb_nhd_gridmet)
Linking to GEOS 3.9.1, GDAL 3.2.1, PROJ 7.2.1; sf_use_s2() is TRUE
* start target p1_drb_nhd_gridmet
x error target p1_drb_nhd_gridmet
* end pipeline: 0.94 seconds
Error : _paths_ missing files: 1_fetch/in/drb_climate_2022_06_14.nc
x Problem with the pipeline.
i Show errors: tar_meta(fields = error, complete_only = TRUE)
i Learn more: https://books.ropensci.org/targets/debugging.html
Error:
! problem with the pipeline.
>
```

lekoenig:
For now, we'll plan to store the gridmet data (full DRB, aggregated to NHDPlusv2 catchments; ~15 GB) on caldera, load the data via a plain targets "file" format, and subset to the desired COMIDs within our pipeline. If running the pipeline on a local clone of the repo, you'd have to download the gridmet data from our S3 bucket and manually place it in 1_fetch/in though.

lekoenig:
Closing this for now as addressed in #141. We can re-open this issue if we decide to pull from S3 down the line.

PR #133, opened 2022-06-16
[#131] s3 gridmet

jsadler2:
moving gridmet into pipelin via S3


lekoenig:
Closing this draft PR for now since the nhd gridmet driver data are brought into the pipeline via changes in #141. We can revisit this PR if we decide to use S3.

Issue #122, opened 2022-06-13
Run model with medium observed sites


galengorski:
I am going to add a model run for medium observed sites, see if I can produce results and troubleshoot issues while Jeff is still around :(

galengorski:
I am first trying to run the 0_baseline_LSTM model before I add another model with the medium observed sites. When I run `tar_make(p2a_metrics_files)` I get the following error:

![image](https://user-images.githubusercontent.com/32860978/173439055-152d19b4-88c2-40fa-a178-ac317cae9077.png)

it looks like an environment issue to me with the 'snakemake' not found errors. This seems like it would be solved by running on tallgrass. I haven't worked with github and tallgrass together, @jsadler2, I'm wondering if you have suggestions on what a good workflow would be. Should I push the branch that I am working on now on my local up to the github repo and then pull it into tallgrass...?

jsadler2:
Yep. Pushing from local to GH and then pulling from GH to Tallgrass works great.

jsadler2:
You should clone your repo, then checkout your fork (though I don't see your fork listed, so I'll show what it would look like for mine). You can do this on the commandline. I'd recommend that over Rstudio.

```
cd /caldera/projects/usgs/water/impd/pump/drb-do
git clone git@github.com:jsadler2/drb-do-ml.git drb-do-ml-ggorski  # putting a suffix on the clone since it's a shared folder
git checkout [your branch]
```

galengorski:
@lekoenig , here is a little update on this issue. I am running into some errors that I think are associated with the dimensions of the new NHD data compared to the older NHM inputs.
- I am trying to run the 0_baseline_model with the new NHD fabric data for the well observed sites 
- I am able to run the `p2a_well_obs_data` target and it produces the output `well_obs_io.zarr`
- One change I had to make was to change the input x variables from column names derived from NHM input to those derived from NHD input
  - The old column names:  
    - x_vars: ['seg_ccov', 'seg_rain', 'seg_slope', 'seg_tave_air', 'seginc_swrad', 'hru_slope', 'hru_aspect', 'hru_elev', 'hru_area', 'hru_percent_imperv', 'covden_sum', 'covden_win', 'soil_moist_max']
  - The new column names:
    - x_vars: ['CAT_CNPY11_BUFF100','pr','SLOPE','tmmx','srad','CAT_BASIN_SLOPE','CAT_BASIN_AREA','CAT_IMPV11']
[...]

PR #139, opened 2022-06-17
Med obs model

galengorski:
Ok this is ready for review now. I realized that the error that I was getting was because I wasn't getting the model to retrain, so I have explicitly run the `train_model` step in the snakemake file with the `-f` forcerun command. I'm sure there is a more elegant and flexible solution to this.


2a_model/src/models/0_baseline_LSTM/config.yml
```diff
@@ -1,6 +1,8 @@
 exp_name: "0_baseline_LSTM"
 
-x_vars: ['seg_ccov', 'seg_rain', 'seg_slope', 'seg_tave_air', 'seginc_swrad', 'hru_slope', 'hru_aspect', 'hru_elev', 'hru_area', 'hru_percent_imperv', 'covden_sum', 'covden_win', 'soil_moist_max']
+site_set: "well_obs"
+
+x_vars: ['pr','SLOPE','tmmx','srad','CAT_BASIN_SLOPE','CAT_ELEV_MEAN','CAT_BASIN_AREA','CAT_IMPV11', 'CAT_CNPY11_BUFF100','CAT_TWI']
```
lekoenig, line 5: Ahh! This makes a lot more sense now. We should probably add a comment somewhere in `2a_model.R` (maybe above `p2a_med_obs_data` and `p2a_well_obs_data`) that reminds a user that `x_vars` needs to be updated in this config file if any variables are added/omitted.

2a_model/src/models/Snakefile_base.smk
```diff
@@ -100,7 +101,7 @@ rule make_predictions:
     input:
         "{outdir}/prepped.npz",
         "{outdir}/nstates_{nstates}/nep_{epochs}/rep_{rep}/train_weights/",
-        "../../../out/well_obs_io.zarr",
+        "../../../out/"+site_set+"_io.zarr",
```
lekoenig, line 104: I like this addition to make the "site set" more flexible.

lekoenig (approved):
Thanks, @galengorski! I haven't run or commented on the changes to the snakemake files, but this looks good to me!

lekoenig:
One more note, @galengorski - do you mind linking issue #122 to this PR so that it gets automatically closed whenever you merge? You can do that by clicking the 'development' cog in the upper right corner of this PR page and selecting the relevant issue.

2a_model/src/models/0_baseline_LSTM/config.yml
galengorski: Good idea! I added a comment in `2a_model` above `p2a_well_obs_data`

2a_model.R
```diff
@@ -91,26 +96,31 @@ p2a_targets_list <- list(
...
     # paths are relative to 2a_model/src/models
     list(list(model_id = "0_baseline_LSTM",
-              snakefile_dir = "0_baseline_LSTM",
-              config_path = "0_baseline_LSTM/config.yml"),
+               snakefile_dir = "0_baseline_LSTM",
+               config_path = "0_baseline_LSTM/config.yml"),
          # the 1_ models use the same model and therefore
          # the same Snakefile as the 0_baseline_LSTM run
-         list(model_id = "1_metab_multitask",
+         # list(model_id = "1_metab_multitask",
```
jsadler2, line 112: Why did you comment these out? Did you try them and they didn't work?

2a_model.R
```diff
@@ -119,25 +129,34 @@ p2a_targets_list <- list(
...
+    system(stringr::str_glue("snakemake {output_feather_file} -s {snakefile_path} --configfile {config_path} -j -f"))
+    
     # Then touch all of the existing files. This makes the weights "up-to-date"
     # so snakemake doesn't train the models again
-    system(stringr::str_glue("snakemake -s {snakefile_path} --configfile {config_path} -j --touch"))
+    #system(stringr::str_glue("snakemake -s {snakefile_path} --configfile {config_path} -j --touch"))
 
     # then run the snakemake pipeline to produce the predictions and metric files
-    system(stringr::str_glue("snakemake -s {snakefile_path} --configfile {config_path} -j --rerun-incomplete"))
+    system(stringr::str_glue("snakemake -s {snakefile_path} --configfile {config_path} -j -f"))
```
jsadler2, line 159: If I'm understanding this correctly, one problem with this approach is that it will rerun the full Snakemake pipeline the `p2a_metrics_files` target needs to be run. For example, if I clone this onto my machine, it won't have the metrics file for exp `0`, so targets will trigger the rebuilding of that branch of the `p2a_metrics_files`. Then it will 
 - build the prepped data file, which good (line 142)
 - retrain the `rep_0` model (line 143)
 - make the `rep_0` preds (line 144)
 - build the overall metrics files (line 159) which would include making all of the predictions (unless they are already there, like `rep_0` (line 144))

jsadler2: I think the reason you added these lines in here was to force a retraining of the model because the previous trained model weights were different than the new, NHD ones. Is that right? I think a cleaner way to do that would be to just delete the previous model weights and run this target as it was written before. That should trigger a retraining of all of the model replicates etc. Did you try that?

2a_model/src/models/Snakefile_base.smk
```diff
@@ -29,7 +30,7 @@ rule as_run_config:
 
 rule prep_io_data:
     input:
-        "../../../out/well_obs_io.zarr",
+        "../../../out/"+site_set+"_io.zarr",
```
jsadler2, line 33: Maybe personal preference, but I think this is a little be cleaner. Using Python's string formatting. That way you don't have to open and close the quotes and I think it's a little less error prone since it's harder to accidentally add in an extra space.
```suggestion
        f"../../../out/{site_set}_io.zarr",
```

2a_model/src/models/Snakefile_base.smk
jsadler2, line 104:
```suggestion
        f"../../../out/{site_set}_io.zarr",
```

2a_model/src/models/Snakefile_base.smk
```diff
@@ -174,7 +175,7 @@ def get_grp_arg(wildcards):
  
 rule combine_metrics:
      input:
-          "../../../out/well_obs_io.zarr",
+          "../../../out/"+site_set+"_io.zarr",
```
jsadler2, line 178:
```suggestion
          f"../../../out/{site_set}_io.zarr",
```

2a_model/src/models/visualize_models.smk
```diff
@@ -1,9 +1,11 @@
 from model_plots import plot_obs_preds
 
+site_set = config['site_set']
+
 rule make_obs_preds_plots:
     input:
         pred_file="{outdir}/nstates_{nstates}/nep_{epochs}/rep_{rep}/preds.feather",
-        obs_file="../../../out/well_obs_targets.zarr",
+        obs_file="../../../out/"+site_set+"_io.zarr",
```
jsadler2, line 8:
```suggestion
        obs_file="../../../out/{site_set}_io.zarr",
```

jsadler2 (changes requested):
Nice work, @galengorski. 

I have a few comments and I wanted to touch base about your changes to the `2a_metrics_files` target before merging.

2a_model.R
galengorski: Ah, just so I could specifically run the baseline models, I will uncomment them

2a_model.R
galengorski: Oh yeah, good point! I think this makes sense, I am going through and testing it to make sure it works. One odd behavior that I am running into: When I got through and delete the weights `out/models/0_baseline_LSTM/nstates_10/nep_100/rep_0/train_weights`, then rerun `p2a_metrics_files` using the original set of snakemake commands (`prepped.npz`,`touch`, and `rerun incomplete`), the model produces a new `train_weights` file for all model ids except 0_baseline_LSTM. Even if I delete all the model output, everything within `out/models/0_baseline_LSTM/nstates_10`, I still don't get the model to retrain

PR #137, opened 2022-06-17
131a s3 gridmet small

jsadler2:
parallel to #133 for testing.

PR #141, opened 2022-06-17
Add NHDv2 gridmet targets to pipeline

lekoenig:
This PR adds two targets to the pipeline, including a file target that represents the NHD-aggregated gridmet driver data for the full DRB (`p1_drb_nhd_gridmet`) and another that subsets the gridmet data to those COMIDs where we have observations (`p2_met_data_at_obs_sites`). 

Here's a preview of the subset data:
```
> tar_load(p2_met_data_at_obs_sites)
> head(p2_met_data_at_obs_sites)
# A tibble: 6 x 10
    COMID time                 tmmx  tmmn    pr  srad    vs  rmax  rmin     sph
    <dbl> <dttm>              <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>   <dbl>
1 4492036 1978-12-31 19:00:00  61.9  47.8 0.481  35.8  9.57  86.1  58.5 0.00628
2 4495680 1978-12-31 19:00:00  59.3  48.8 0.632  35.6  7.70  82.8  66.0 0.00686
3 4495656 1978-12-31 19:00:00  58.4  48.0 0.634  35.7  7.63  84.2  66.9 0.00678
4 4784831 1978-12-31 19:00:00  59.9  48.9 0.610  35.5  7.77  82.2  65.5 0.00690
5 4782619 1978-12-31 19:00:00  58.1  48.3 0.625  35.5  7.61  84.8  67.5 0.00679
6 4782585 1978-12-31 19:00:00  57.3  46.7 0.610  36.0  7.57  88.8  69.0 0.00673
>
```

These targets were taken from #137 and modified a bit. The gridmet driver data are treated as a plain 'file' format target as opposed to a target that integrates with S3 and has `repository = aws`. This assumes that the DRB-scale gridmet data are available in 1_fetch/in.

### DIFF
--- a/1_fetch.R
+++ b/1_fetch.R
@@ -300,6 +300,17 @@ p1_targets_list <- list(
     read_csv(p1_ref_gages_manual_csv, col_types = cols(.default = "c")) %>%
       mutate(id = str_replace(provider_id, "USGS-","")) %>%
       relocate(id, .after = provider_id)
+  ),
+  
+  # Read in meteorological data aggregated to NHDPlusV2 catchments for the 
+  # DRB (prepped in https://github.com/USGS-R/drb_gridmet_tools). Note that
+  # the DRB met data file must be stored in 1_fetch/in. If working outside
+  # of tallgrass/caldera, this file will need to be downloaded from the project
+  # S3 bucket and manually placed in 1_fetch/in.
+  tar_target(
+    p1_drb_nhd_gridmet,
+    "1_fetch/in/drb_climate_2022_06_14.nc",
+    format = "file"
   )
 
 )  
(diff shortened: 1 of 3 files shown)


lekoenig:
> Just want to make sure I understand: the assert call within subset_nc_to_comids is to make sure that you are only filtering out 4781767, which is the zero area segment, is that correct?

Yeah, exactly. NHDv2 COMID's with `AREASQKM == 0` will not have associated catchment areas and therefore won't have met driver data, either.

To aggregate the gridmet data, Jeff used catchment polygons that are created in the drb-network-prep repo (`p1_nhdv2_catchments_sf` [here](https://github.com/USGS-R/drb-network-prep/blob/main/1_fetch.R)). Within that pipeline we now omit any zero-area COMID's before downloading the catchments. So this assert statement shouldn't be necessary, but we decided to leave it in for now as a sort of flag.

galengorski:
Awesome, thanks for the explanation, looks good then!

Issue #148, opened 2022-08-16
Should  `2a_model.R` use med-obs or well-obs sites (but not both)?


lekoenig:
Currently, `2a_model.R` builds separate targets and input files for the "well-observed" sites and the "medium-observed" sites. This makes the file more difficult to interpret/remember to update each relevant target, and might be over-engineered for what we need. Will we always want _both_ the medium-observed and well-observed outputs, or is it more likely that we'll just choose one and train the model using that dataset?

What if we created a new target called `p2a_observed_sites` or something like that that just serves as a toggle for which set of sites we're interested in, e.g.:

```r
# p2_well_observed_sites could be replaced with p2_med_observed_sites below
tar_target(
p2a_observed_sites
   p2_well_observed_sites
),
```

This way, we would update the downstream targets to depend on `p2a_observed_sites` instead of either `p2_well_observed_sites` or `p2_med_observed_sites`, and we would create just one version of the input/output zarr data store depending on which set of sites we're interested in.

lekoenig:
^ One place where I'm not sure what the impact of this proposed change would look like is `p2a_metrics_files`, so I'd also appreciate specific thoughts on that target.

lekoenig:
Assuming we're happy with the medium-observed sites (see #122) and if we don't need to separately define medium- and well-observed sites for the analysis or a paper, we could also just create an object called `n_days_well_obs <- 100` in `_targets.R` and then use that object in place of `300` when defining `p2_well_observed_sites` and then just get rid of all targets related to "medium observed" data. 

I'm in favor of this change because it would simplify `2a_model.R` and reduce the chance a coding error gets made, e.g. if we forget to update one reference from "well-observed" to "med-observed". Do you agree, @galengorski, @jsadler2? Any reason to keep both?

galengorski:
I think making the complete switch over to medium observed sites makes sense, it looks like it improves the performance and I can't think of a reason that we would want to be able to toggle back to the well observed sites.

PR #149, opened 2022-08-16
Add site map to pipeline

### LINKED ISSUE #71: Create map of study domain
We'll need a map of the lower DRB for the paper, presentations

lekoenig:
This PR adds a draft site map to the pipeline and also includes some code formatting changes that aren't related to the site map but try to adopt a more consistent and readable code style throughout the repo. 

I created a version of this map for our JASM presentation and am adding it here as a starting place for a site map figure to include in a publication. To color-code the sites by site type, I added a new target `p2a_site_splits` (which replaces an older version called `p3_well_observed_site_data`). For the JASM presentation I did not have a category called "train/val" which represents those training sites that have data within the validation date range. I'm not sure what our preference is for displaying those "time split" sites as opposed to just the spatial splits, so let me know if you have thoughts on which site types we should display. 

Here's a preview of the output from `p3_site_map_png`:
 
do_site_map

Note that this site map currently only includes the "well-observed" sites because I did not want to modify the targets `p2a_trn_sites`, `p2a_trn_val_sites`, and `p2a_trn_sites_w_val_data` which are used to create model input files in `2a_model.R` (see #148). 

Closes #71


galengorski (approved):
Looks good to me, I don't see anything that will affect the modeling process in `2a_model.R` (which is where I mostly focused). I like the map too!

PR #154, opened 2022-08-24
Omit medium-observed targets

### LINKED ISSUE #148: Should  `2a_model.R` use med-obs or well-obs sites (but not both)?

lekoenig:
This PR makes the following code changes to simplify the pipeline:

- add an object called `min_obs_days` to `_targets.R`. `min_obs_days` is used to define the minimum number of observation-days a site should have before it's considered "well-observed" and used in the model. `min_obs_days` replaces `"300"` in `p2_well_observed_sites` and therefore serves as an input that can be modified by the user. 
- define `min_obs_days <- 100` so that our "well-observed" site targets now reflect our decision to use what we were previously calling "medium-observed" sites (see #122 for more detail)
- omit targets that refer to "med-observed" sites in `2_process.R` and `2a_model.R`

To maintain consistency with our previous model runs using the medium-obs sites, `p2_well_observed_sites` only looks for NWIS daily or instantaneous sites with at least 100 days, although some of our discrete sites also contain data from 100+ days.

Closes #148

### DIFF
--- a/2_process.R
+++ b/2_process.R
@@ -140,34 +140,15 @@ p2_targets_list <- list(
    }
  ),
  
- # make list of "moderately-observed" sites
- tar_target(
-   p2_med_observed_sites,
-   p2_sites_w_segs %>%
-     filter(count_days_nwis >= 100) %>%
-     pull(site_id)
- ),
- 
- # filter p1_reaches_sf to segments with "well-observed" sites
- tar_target(   
-   p2_med_observed_reaches,
-   {
-     med_obs_reach_ids <- p2_sites_w_segs %>%
-       filter(site_id %in% p2_med_observed_sites) %>% 
-       pull(COMID)
-     p1_nhd_reaches_sf %>% filter(COMID %in% med_obs_reach_ids)
-   }
- ),
- 
  # Estimate daily (normalized) max-light
  tar_target(
    p2_daily_max_light,
    { 
-     calc_seg_light_ratio(p2_med_observed_reaches, 
+     calc_seg_light_ratio(p2_well_observed_reaches, 
                           start_date = earliest_date, 
                           end_date = latest_date)
    },
-   pattern = map(p2_med_observed_reaches)
+   pattern = map(p2_well_observed_reaches)
  ),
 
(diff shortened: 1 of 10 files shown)


lekoenig:
The changes in this PR don't include any edits to the model config files (e.g. `site_set`), but probably should. Should I just update model config files to use "well-obs" for `site_set`?

jsadler2 (approved):
This looks good to me.

We have the old "well-observed" results to look back on if we want, correct?

lekoenig:
Jeff and I decided that the simplest approach was to hard-code in the path of the well-obs zarr files (I added a note in `2a_model.R` letting the user know about this dependency) and to remove the variable `site_set` from each of the model config files.

@jsadler2, do you mind taking a quick look at the snakefile/cfg files? I haven't tried running this yet because I want to first resolve the issue I'm having running the pipeline with `tmmn` on tallgrass.

jsadler2 (changes requested):
This looks good. Thanks, @lekoenig 

I do suggest one additional comment saying where our well observed threshold is defined (_targets.R). 

Also, when I tried to run this, I found that the 2a_model/src/models/visualize_models.smk file has `site_set` still in it. So that needs to be changed.

jsadler2 (approved):
👍

Issue #153, opened 2022-08-24
Rerun baseline LSTM with tmmn and tmmx


lekoenig:
In our meeting on 8/23/2022 we agreed that we wanted to use both daily min and max temperature from gridmet as input features (rather than `tmmx` alone).

PR #155, opened 2022-08-30
run models with tmmn and tmmx

galengorski:
This PR is in reference to #153 , running all four models with tmmx and tmmn. The PR updates the config files and the weights from the model runs


2a_model/src/models/1_metab_multitask/2b_multitask_do_gpp.yml
```diff
@@ -2,7 +2,7 @@ exp_name: "1b_multitask_do_gpp"
 
 site_set: "well_obs"
```
lekoenig, line 3: Not related to "tmmx" or "tmmn", but is there a reason we use "well_obs" for `site_set` here, but "med_obs" for the others?

lekoenig (approved):
Looks good, @galengorski! I added just one (unrelated) question about the model config files. And I'm curious - can you tell if adding `tmmn` had any impact on the model performance?

lekoenig:
> And I'm curious - can you tell if adding tmmn had any impact on the model performance?

I now see that you've included a few plots in #152 😊

jsadler2 (approved):
Looks good to me.

PR #156, opened 2022-09-07
add functional performance plotting scripts

galengorski:
This has three scripts for generating functional performance plots:

1. `it_functions.py` is a script with generic functions for calculating IT metrics
2. `do_it_functions.py` is a wrapper script which calls in the model inputs and outputs and calculates the it metrics for each site and do variable
3. `do_functional_performance_calcs_plotting.py` calls functions from `do_it_functions.py` for the calculations and plotting


galengorski:
I incorporated edits from @jsadler2 (#162), and added: 
1. functionality to run a subset of sites
2. `get_max_it_df` function now writes the results to a csv
3. a snakemake file for running the calculations

lekoenig:
Awesome 😎, thanks @galengorski! I don't have any comments on this PR so will defer to you and @jsadler2 to agree on the changes and merge.

jsadler2 (changes requested):
I have a few comments. I think at a high level, we need to decide how best to plug this into the overall pipeline. I have one idea for doing that that I've put in the inline comments.

scratch/Functional_Performance/Snakefile_functional_performance.smk
```diff
@@ -0,0 +1,31 @@
+import sys
+sys.path.insert(0, 'C:\\Users\\ggorski\\OneDrive - DOI\\USGS_ML\\DO\\drb-do-ml\\scratch\\Functional_Performance\\src')
```
jsadler2, line 2: I'm curious why you changed this to an absolute path. I think we will need the relative path here. Did `"src"` not work?

jsadler2 (commented):
Looks good. I do have one question about the path you are adding to the system path.

For the `do_functional_performance_calcs_plotting.py` it's probably going to be easiest to do a `git rm` on that file and then commit that change. 

E.g.,
```
git rm do_functional_performance_calcs_plotting.py
git commit -m "[#162] remove functional perf plotting script"
```

Now that I'm looking at this though, I'm wondering if we should keep any part of the plotting script. The functional perf. calculations are now happening in the snakefile, so I think those parts should come out, but do you think it would be good to leave the plotting code at least for our reference?

scratch/Functional_Performance/Snakefile_functional_performance.smk
galengorski: This is mostly for my own sanity, I am working on two sets of it functions, one for DO and one for Estuary Salinity, they rely on the same code but the wrapper functions are slightly different, so I just want to make sure that I am loading the correct module

galengorski:
> Looks good. I do have one question about the path you are adding to the system path.
> 
> For the `do_functional_performance_calcs_plotting.py` it's probably going to be easiest to do a `git rm` on that file and then commit that change.
> 
> E.g.,
> 
> ```
> git rm do_functional_performance_calcs_plotting.py
> git commit -m "[#162] remove functional perf plotting script"
> ```
> 
> Now that I'm looking at this though, I'm wondering if we should keep any part of the plotting script. The functional perf. calculations are now happening in the snakefile, so I think those parts should come out, but do you think it would be good to leave the plotting code at least for our reference?

Sounds good, I'll delete the file. We could put a version in the `3_visualize/src` directory if we wanted it for reference.

jsadler2 (approved):
+1

PR #157, opened 2022-09-08
5 replicates nhd

jsadler2:
5 replicates for the nhd model


lekoenig:
@jsadler2 Is there value in still committing/merging these trained weights with 5 model reps? This is the "slim" nhd model with minimal input features, right?

Issue #160, opened 2022-09-09
Decide whether to define val time period in `_targets.R`


lekoenig:
Currently, the time period used for model validation is defined in two places, `_targets.R` and the model [yml](https://github.com/USGS-R/drb-do-ml/blob/0db03f372ee84824a51459cb96e33943842f1d5b/2a_model/src/models/config_base.yml#L18) file. These two locations also currently contain different time periods, so we should simplify how we define the val times.

One option is to just omit the val times from `_targets.R`. Those val times do get used in [`2a_model.R`](https://github.com/USGS-R/drb-do-ml/blob/0db03f372ee84824a51459cb96e33943842f1d5b/2a_model.R#L59) and are used to create visuals. One downside I see with omitting these variables from the targets file is that in this pipeline `_targets.R` serves as a config file, at least for the data processing steps. I think we can introduce confusion by defining variables separately in the model config files, although a comprehensive README for running/modifying the modeling steps would help with that (see [this](https://github.com/USGS-R/drb-do-ml/issues/146) issue).

PR #163, opened 2022-10-03
Build model config files within targets pipeline

lekoenig:
This PR addresses #160, which notes that we currently define `val_times` in two separate places, in `_targets.R` and in `2a_model/src/models/config_base.yml` (and right now the `val_times` differ between these two places). 

For purposes of interpretability and avoiding bugs, I think we should try to avoid situations where we're defining model parameters in multiple places. We could omit `val_times` from `_targets.R`, although those variables currently get passed to visualization/plotting functions. 

`_targets.R` acts as a config file for the rest of the data-model pipeline, so I've also wondered whether we should just define our model parameters there, and have a target that writes the config files to a yml. This way, `_targets.R` is a kind of one-stop shop for all model inputs and parameters. I've implemented this for the baseline model config file in this draft PR. I would appreciate your ideas, @jsadler2 and @galengorski, on 1) whether this is useful, overkill, both(?); 2) if useful, one question I would have is whether the block style formatting (see `val_sites` and `test_sites` below) would work. I don't know how I can write the parameters using the 'flow-style' sequence formatting (e.g. `val_sites = ["x","y","z"]`) that's currently included in the config files.

Here's what the output config would look like for the baseline model:

```py
out_dir: "../../../out/models"
seed: False
num_replicates: 1
trn_offset: 1.0
tst_val_offset: 1.0
early_stopping: False
train_start_date:
  - '1980-01-01'
train_end_date:
  - '2014-10-01'
val_start_date:
  - '2014-10-01'
val_end_date:
  - '2015-10-01'
test_start_date:
  - '2015-10-01'
test_end_date:
  - '2022-10-01'
validation_sites:
  - "01472104"
  - "01473500"
  - "01481500"
test_sites:
  - "01475530"
  - "01475548"
epochs: 100.0
hidden_size: 10.0
dropout: 0.2
recurrent_dropout: 0.2
finetune_learning_rate: 0.01
```

### DIFF
--- a/2a_model.R
+++ b/2a_model.R
@@ -87,13 +87,22 @@ p2a_targets_list <- list(
       sf::st_as_sf(., coords = c("lon","lat"), crs = unique(.$epsg))
   ),
   
-  ## WRITE MODEL CONFIG FILES ##
+  ## WRITE MODEL CONFIGURATION FILES ##
   # Write baseline model config file using inputs and parameters defined 
   # in the _targets.R file.
   tar_target(
     p2a_config_base_yml,
     write_config_file(cfg_options = base_config_options,
                       fileout = "2a_model/src/models/config_base.yml"),
+    format = "file"
+  ),
+  
+  # Write model config file for 0_baseline_LSTM
+  tar_target(
+    p2a_config_baseline_LSTM_yml,
+    write_config_file(cfg_options = model_config_options,
+                      fileout = "2a_model/src/models/0_baseline_LSTM/config.yml",
+                      exp_name = "0_baseline_LSTM"),
     format = "file"
   ),
   
--- a/2a_model.R
+++ b/2a_model.R
@@ -87,29 +87,13 @@ p2a_targets_list <- list(
       sf::st_as_sf(., coords = c("lon","lat"), crs = unique(.$epsg))
   ),
   
-  # Write baseline model config file
+  ## WRITE MODEL CONFIG FILES ##
+  # Write baseline model config file using inputs and parameters defined 
+  # in the _targets.R file.
   tar_target(
     p2a_config_base_yml,
-    write_base_config_file(fileout = "2a_model/src/models/config_base_test.yml", 
-                           model_save_dir = "../../../out/models", 
-                           seed = seed, 
-                           n_reps = n_model_reps, 
-                           trn_offset = trn_offset, 
-                           tst_val_offset = tst_val_offset, 
-                           early_stopping = early_stopping, 
-                           epochs = epochs, 
-                           hidden_size = hidden_size, 
-                           dropout = dropout,
-                           recurrent_dropout = recurrent_dropout, 
-                           finetune_learning_rate = finetune_learning_rate,
-                           val_sites = val_sites, 
-                           test_sites = tst_sites,
-                           train_start_date = train_start_date, 
-                           train_end_date = train_end_date, 
-                           val_start_date = val_start_date, 
-                           val_end_date = val_end_date,
-                           test_start_date = test_start_date, 
-                           test_end_date = test_end_date),
+    write_config_file(cfg_options = base_config_options,
+                      fileout = "2a_model/src/models/config_base.yml"),
     format = "file"
   ),
   
--- a/2a_model.R
+++ b/2a_model.R
@@ -88,8 +88,7 @@ p2a_targets_list <- list(
   ),
   
   ## WRITE MODEL CONFIGURATION FILES ##
-  # Write baseline model config file using inputs and parameters defined 
-  # in the _targets.R file.
+  # Write base config file using inputs and parameters defined in _targets.R
   tar_target(
     p2a_config_base_yml,
     write_config_file(cfg_options = base_config_options,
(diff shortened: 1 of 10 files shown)


galengorski:
@lekoenig , I generally like this idea, particularly for making sure we only specify parameters in a single location. So this would be a target to create the `config_base.yaml` file found in `2a_model/src/models` and then each model would still have it's own model specific `config.yaml` file where the variables and lambdas are defined, which would be edited manually?

lekoenig:
> So this would be a target to create the config_base.yaml file found in 2a_model/src/models and then each model would still have it's own model specific config.yaml file where the variables and lambdas are defined, which would be edited manually?

@galengorski the idea would be that we create targets that write each of the yaml files we need (to avoid any manual edits to any of the model cfg files). I added this baseline cfg as an example to see if it's something we want to implement and/or if this format will work in the snakemake workflow.

jsadler2 (commented):
Like I said the mtg, I think this is a really smart addition/change, @lekoenig.

One high-level suggestion is that think it would be more efficient/clearer if we just had a list of this in the `_targets.R` file. Then we don't have to write all these things out twice (once in the `_targets.R` file and once in the `write_base_config_file` function. You could just read in the list from the `_targets.R` file and write that to the yaml. Does that make sense? 

I tried to explain more in the inline comments below.

2a_model.R
```diff
@@ -85,7 +86,32 @@ p2a_targets_list <- list(
...
+                           recurrent_dropout = recurrent_dropout, 
+                           finetune_learning_rate = finetune_learning_rate,
+                           val_sites = val_sites, 
+                           test_sites = tst_sites,
+                           train_start_date = train_start_date, 
+                           train_end_date = train_end_date, 
+                           val_start_date = val_start_date, 
+                           val_end_date = val_end_date,
+                           test_start_date = test_start_date, 
+                           test_end_date = test_end_date),
```
jsadler2, line 112:
```suggestion
                           model_save_dir = "../../../out/models", 
                          config_options = base_config_options),
```

2a_model/src/write_model_config_files.R
```diff
@@ -0,0 +1,104 @@
...
+write_base_config_file <- function(fileout, model_save_dir, 
+                                   seed = FALSE, n_reps, 
+                                   trn_offset, tst_val_offset, 
+                                   early_stopping = FALSE, 
+                                   epochs, hidden_size, dropout,
+                                   recurrent_dropout, finetune_learning_rate,
+                                   val_sites, test_sites,
+                                   train_start_date, train_end_date, 
+                                   val_start_date, val_end_date,
+                                   test_start_date, test_end_date){
```
jsadler2, line 51:
```suggestion
write_base_config_file <- function(fileout, model_save_dir, 
                                 config_options){
```

_targets.R
```diff
@@ -76,16 +79,32 @@ min_obs_days <- 100
...
+trn_offset <- 1
+tst_val_offset <- 1
+epochs <- 100
+hidden_size <- 10
+dropout <- 0.2
+recurrent_dropout <- 0.2
+finetune_learning_rate <- 0.01
+# random seed for training False==No seed, otherwise specify the seed
+seed <- FALSE
+early_stopping <- FALSE
```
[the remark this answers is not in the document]
jsadler2: You'd of course want to add `train_start_date`, `val_sites` and `tst_sites` too, but I couldn't add those to the suggestion b/c those lines weren't changed in this PR.

lekoenig:
Thanks for your input and suggestions, Galen and Jeff! I've made some edits to the function `write_config_file()` based on Jeff's suggestions, which also helped make this function more general 🎉. I ended up including a code chunk for each model in `_targets.R` to create an analogous list that contains the cfg options for that model. I think this works fine, but I'm open to any suggestions for making `_targets.R` easy to interpret and know what needs to be updated/changed for a given model run. 

Tagging @jsadler2 for a review, which could be high-level, e.g. do the updated cfg files look right? does the model run OK with these changes?

lekoenig:
In our meeting on 10/12, @jsadler2 pointed out that we also need to specify the new yml targets as dependencies to `p2a_metrics_files`. I've done that in https://github.com/USGS-R/drb-do-ml/pull/163/commits/c0ab612980c3bf62d29047130ef6b6ed64f2fdf6 by replacing the hard-coded config file paths in `p2a_model_ids` with the yml targets themselves. This way, a change in any of the cfg files will trigger a rebuild of `p2a_model_ids`, which will then trigger `p2a_metrics_files`.

p2a_metrics_files_dep

_targets.R
```diff
@@ -76,16 +79,85 @@ min_obs_days <- 100
...
+  trn_offset = 1,
+  tst_val_offset = 1,
+  epochs = 100,
+  hidden_size = 10,
+  dropout = 0.2,
+  recurrent_dropout = 0.2,
+  finetune_learning_rate = 0.01,
+  early_stopping = FALSE,
+  # train/val/test split information is defined above:
+  val_sites = val_sites,
```
jsadler2, line 112:
```suggestion
  validation_sites = val_sites, 
```
 `validation` was what was in the config files originally

jsadler2 (changes requested):
Lauren - this looks good overall. I have a couple small suggestions below. 

Also, I got an error when I tried to run it b/c it was looking for `CAT_BASIN_AREA`. I checked and couldn't find it in the Zarr. Is that expected? Also, am I remembering right that you were thinking we should drop that? If that's the case, I guess we'd just need to take it out of the `x_vars_global` vector.

_targets.R
jsadler2: when I ran this, it threw an error, b/c Snakemake was looking for the `validation_sites` config option.

lekoenig:
> Also, I got an error when I tried to run it b/c it was looking for CAT_BASIN_AREA. I checked and couldn't find it in the Zarr. Is that expected?

Yeah, I rebased the branch on my fork yesterday, so if you ran the model this morning CAT_BASIN_AREA would have no longer been downloaded/included in the seg attributes that get saved to the Zarr. I've removed CAT_BASIN_AREA from the vector of variables in `x_vars_global` in https://github.com/USGS-R/drb-do-ml/pull/163/commits/a9da5c3f43be883eab57709947c4c5cc9d3aa976.

_targets.R
lekoenig: Great catch, thanks! I've accepted your suggested fix in https://github.com/USGS-R/drb-do-ml/pull/163/commits/a9da5c3f43be883eab57709947c4c5cc9d3aa976.

_targets.R
```diff
@@ -76,16 +79,85 @@ min_obs_days <- 100
...
+  train_start_date = train_start_date, 
+  train_end_date = train_end_date, 
+  val_start_date = val_start_date, 
+  val_end_date = val_end_date,
+  test_start_date = test_start_date, 
+  test_end_date = test_end_date
+  )
+
+# Define global model parameters for the "baseline" deep learning model
+x_vars_global <- c("pr","SLOPE","tmmx","tmmn","srad","CAT_BASIN_SLOPE","CAT_ELEV_MEAN",
```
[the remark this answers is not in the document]
jsadler2: There may be a case where we decide to change `x_vars` for a particular model. I suggest still putting in the `base_config_options`. If there is a case where we want to change it for one model, we can provide a `x_vars` option in that model's configuration list and it will override the base_config's config option.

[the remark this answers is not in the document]
lekoenig: Oh yeah, good point. I've added `x_vars_global` to the `base_config_options` list in https://github.com/USGS-R/drb-do-ml/pull/163/commits/8274cd1371abfaba6e767865cd7291b2aade0f5d.

jsadler2 (approved):
Looks great.

PR #168, opened 2022-10-20
Plot segment/catchment attributes

lekoenig:
This PR creates plots to quickly visualize the segment/catchment attributes that were downloaded and processed. These plots are meant to help understand the input features and aid feature selection. 

A new function, `plot_nhdv2_attr()`, plots the distribution of attribute values as well as their spatial distribution in the lower DRB. The function is used in a new target in `3_visualize.R` - `p3_nhdv2_attr_png` - which saves those plots to a folder in 3_visualize/out.

Here's an example of the plots that are generated by the code changes in this PR:

CAT_CLAYAVE

### DIFF
--- a/3_visualize/src/plot_nhdv2_attr.R
+++ b/3_visualize/src/plot_nhdv2_attr.R
@@ -1,0 +1,134 @@
+#' @title Plot segment and catchment attributes
+#' 
+#' @description 
+#' This function visualizes each of the NHDPlusv2 attribute variables across
+#' all river segments within the network.
+#' 
+#' @details 
+#' This function was originally developed as part of the drb-inland-salinity-ml 
+#' project:
+#' https://github.com/USGS-R/drb-inland-salinity-ml/blob/synoptic_site_viz/3_visualize/src/plot_nhdv2_attr.R
+#'
+#' @param attr_data data frame containing the processed NHDv2 attribute data; 
+#' must include column "COMID".
+#' @param network_geometry sf object containing the network flowline geometry; 
+#' must include columns "COMID" and "geometry".
+#' @param save_dir character string indicating where the directory where the
+#' output plots should be saved.
+#' @param plot_sites logical; indicates whether or not to plot sampling sites.
+#' @param sites tbl with the sampling sites and columns for the corresponding "COMID".
+#' @param sites_epsg integer indicating the epsg code in the sites table 
+#' (i.e., 4269 for NAD83).
+#'
+#' @returns 
+#' Saves a png file containing a violin plot showing the distribution of each 
+#' NHDv2 attribute variable and returns the file path of the saved file. 
+#' 
+plot_nhdv2_attr <- function(attr_data,
+                            network_geometry,
+                            save_dir,
+                            plot_sites = FALSE, 
+                            sites = NULL, 
+                            sites_epsg = NULL){
+
+  if(plot_sites){
+    if (is.null(sites)){
+      stop('sites must be specified when plot_sites = TRUE')
+    }
+    if (is.null(sites_epsg)){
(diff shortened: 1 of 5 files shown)


3_visualize/src/plot_nhdv2_attr.R
```diff
@@ -0,0 +1,134 @@
...
+        left_join(.,network_geometry[,c("COMID","geometry")],by=c("COMID"="COMID")) %>%
+        sf::st_as_sf() %>%
+        ggplot() + 
+        geom_sf(aes(color=.data[[col_name]]), size = 0.3) + 
+        scale_color_viridis_c(option="plasma") + 
+        theme_bw() + 
+        theme(plot.margin = unit(c(0,0,0,2), "cm"),
+              axis.text.x = element_text(size = 6),
+              legend.title = element_text(size = 10)) +
+        geom_sf(data = sites_sf, size = 0.3)
```
galengorski, line 95: The black dots for the site locations on the map are kind of hard to see, what about a white dot with a black outline? I don't know, pretty minor though

galengorski (approved):
@lekoenig , I think these look great! It will be really helpful to be able to visualize these attributes quickly, and I really like the context of the whole DRB. One thing that I might be useful, either as a separate plot or combined with these is a figure that labels the sites, either on a bar chart or on a map.  For example, in the %Clay figures that you added in the PR, it's difficult to tell which site is the one with a really high %Clay, and if we are trying to connect attributes to performance, it might be nice to have that information readily available from these plots (or something like them). I also added a small comment about the color of the sites on the map, so they are easier to pick out, but otherwise this is great!

3_visualize/src/plot_nhdv2_attr.R
lekoenig: Yeah, I know what you mean! As a first cut, I tried to make these points a little bit bigger and also added a bit of transparency to the river flowlines. I think this makes the site locations more visible, let me know what you think about that.

lekoenig:
> One thing that might be useful, either as a separate plot or combined with these is a figure that labels the sites, either on a bar chart or on a map. 

Good point! I opted not to include site labels on the map because the map was really cluttered since the sites are concentrated in a relatively small area. Instead, I added an optional argument, `label_sites`, to the `plot_nhdv2_attr()` function that adds site labels to the red points on the violin plot. Here's the same %clay example (the user can also adjust the size of the labels):


CAT_CLAYAVE

So it adds some clutter but as a relatively quick cut, might help quickly ID which sites have relatively higher or lower values of a given attribute. Do you think this helps?

3_visualize/src/plot_nhdv2_attr.R
galengorski: > > One thing that might be useful, either as a separate plot or combined with these is a figure that labels the sites, either on a bar chart or on a map.
> 
> Good point! I opted not to include site labels on the map because the map was really cluttered since the sites are concentrated in a relatively small area. Instead, I added an optional argument, `label_sites`, to the `plot_nhdv2_attr()` function that adds site labels to the red points on the violin plot. Here's the same %clay example (the user can also adjust the size of the labels):
> 
> CAT_CLAYAVE
> 
> So it adds some clutter but as a relatively quick cut, might help quickly ID which sites have relatively higher or lower values of a given attribute. Do you think this helps?

Yeah I see what you mean, I like the violin plot point labeling option. Even if it does add some clutter, the default could be no labels and we could just flip the switch to check and compare quickly

lekoenig:
> Even if it does add some clutter, the default could be no labels and we could just flip the switch to check and compare quickly

Sounds good, I agree with that. I see that you 'approved' the PR - should I go ahead and merge?

galengorski:
Sounds good to me

lekoenig:
Thanks for the review, @galengorski!

Issue #159, opened 2022-09-09
Clean up/finalize baseline LSTM


lekoenig:
This issue captures several tasks that are needed to clean up/finalize the baseline LSTM:

- [x] Add `tmmn` to list of meteorological drivers #153
- [x] Check date formatting for gridmet data
- [ ] Incorporate gridmet data in targets pipeline #151 
- [x] Add additional segment/catchment attributes at NHDPlusv2-scale #51
- [x] `inputs_and_outputs_med_obs` in `2a_model.R` has redundant COMID columns
- [x] Simplify pipeline by omitting targets for _both_ well-observed and medium-observed sites #148 
- [x] Decide whether provisional data will be included in the observations used for model training #135 
- [x] Val times are currently defined in two places, and the values differ between those locations #160 
- [ ] Flesh out README to describe the modeling inputs and steps, including how to modify and run the model #146

Issue #146, opened 2022-07-28
Add model instructions to README


lekoenig:
Ahh! This makes a lot more sense now. We should probably add a comment somewhere in `2a_model.R` (maybe above `p2a_med_obs_data` and `p2a_well_obs_data`) that reminds a user that `x_vars` needs to be updated in this config file if any variables are added/omitted.

_Originally posted by @lekoenig in https://github.com/USGS-R/drb-do-ml/pull/139#discussion_r932512432_

lekoenig:
Jotting down some items we might want to include:

- [ ] use of git submodule to run river-dl
- [ ] how to run the model without making any changes
- [ ] where to modify model config inputs
- [ ] use of committed weights vs retraining a model

PR #169, opened 2022-10-21
Format model inputs in p2a_model.R

lekoenig:
This PR addresses part of #159, namely that there were redundant COMID columns in the `inputs_and_outputs` data frame in `p2a_well_obs` (e.g. `COMID.x`, `COMID.y`). Here I've added `COMID` as a join key when building `p2a_well_obs` to omit the redundant columns.

I've also added some comments and made formatting edits in both `2a_model.R` and `2a_model/src/model_ready_data_utils.R` to 1) add documentation that will allow us/me to better track the different steps in `2a_model.R`; and 2) format some function documentation to match the code style used throughout the rest of the repo.

### DIFF
--- a/2a_model.R
+++ b/2a_model.R
@@ -3,7 +3,7 @@ source("2a_model/src/write_model_config_files.R")
 
 p2a_targets_list <- list(
 
-  ## PREPARE (RENAME, JOIN) INPUT AND OUTPUT FILES ##
+  ## 1) COMBINE AND FORMAT MODEL-READY INPUTS AND OUTPUTS ##
   # join met data with light input data
   tar_target(
     p2a_met_light_data,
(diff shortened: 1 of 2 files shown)


galengorski (approved):
Hi @lekoenig this looks good to me! I'm wondering if there is a way that we can map the snakemake dependencies per your comments on lines 171-174, I always find it confusing to track down which of the snakefile need changing, but maybe that is just the nature of a complex pipeline.

lekoenig:
> I'm wondering if there is a way that we can map the snakemake dependencies per your comments on lines 171-174, I always find it confusing to track down which of the snakefile need changing, but maybe that is just the nature of a complex pipeline.

I agree and would love to incorporate any suggestions for taking the burden of tracking dependencies out of our heads and into comments and code! Do you have an example of what you mean by mapping the snakemake dependencies? Are you referring to added documentation somewhere?

lekoenig:
My thought (hope?) was to treat the `_targets.R` file as a sort of one-stop-shop location for configuration options that relate to the data + model pipeline. Hopefully this would reduce the cognitive burden of knowing where to go to update data or modeling options (even if you're not familiar with the individual data processing steps in the targets pipeline or all of the modeling steps in the snakemake pipeline). 

So now we include the model configuration options in `_targets.R` and the configuration yaml files get built by the targets pipeline. And I _think_ that individual lines in the `.smk` file should never need to be manually changed, but I'm less familiar with those files and could be wrong about that. 

I suggest we think about this further and make additional changes when we tackle #146. I'll go ahead and merge this PR for now.

PR #174, opened 2022-11-04
Replace hard-coded prediction dates in Snakefile_base.smk

### LINKED ISSUE #171: Check snakefiles for config options?
If we continue to build the model config files in the targets pipeline, we should check that there aren't other config options that are hard-coded in the model snake files. For example, I'm wondering if the dates should be hard-coded in `pred_start_date` and `pred_end_date` below?

```py
rule make_predictions:
    input:
        "{outdir}/prepped.npz",
        "{outdir}/nstates_{nstates}/nep_{epochs}/rep_{rep}/train_weights/",
        f"../../../out/well_obs_io.zarr",
    output:
        "{outdir}/nstates_{nstates}/nep_{epochs}/rep_{rep}/preds.feather",
    run:
        weight_dir = input[1] + "/"
        params.model.load_weights(weight_dir)
        preds = predict_from_arbitrary_data(raw_data_file=input[2],
                                            pred_start_date="1980-01-01",
                                            pred_end_date="2019-01-01",
                                            train_io_data=input[0],
                                            model=params.model, 
                                            spatial_idx_name='site_id',
                                            time_idx_name='date')
        preds.reset_index(drop=True).to_feather(output[0])
```

lekoenig:
This is a small PR to replace hard-coded start and end dates for the predictions step in `Snakefile_base.smk` with dates that get pulled from the config files. I tried running the pipeline with this edit on tallgrass and it seemed to work fine. 

@jsadler2, can you remind me why the `pred_end_date` is equal to the `val_end_date` (as opposed to the `test_end_date`)?

Closes #171.

### DIFF
--- a/2a_model/src/models/Snakefile_base.smk
+++ b/2a_model/src/models/Snakefile_base.smk
@@ -107,8 +107,8 @@ rule make_predictions:
         weight_dir = input[1] + "/"
         params.model.load_weights(weight_dir)
         preds = predict_from_arbitrary_data(raw_data_file=input[2],
-                                            pred_start_date="1980-01-01",
-                                            pred_end_date="2019-01-01",
+                                            pred_start_date = config['train_start_date'],
+                                            pred_end_date = config['val_end_date'],
                                             train_io_data=input[0],
                                             model=params.model, 
                                             spatial_idx_name='site_id',


jsadler2 (approved):

jsadler2:
I was thinking we should do the validation end date just to keep the test set further hidden away

Issue #176, opened 2022-11-22
How many model replicates should we run?


lekoenig:
The model outputs in this repo currently include 1 model replicate (`2a_model/out/0_baseline_LSTM/nstates_10/nep_100/rep0/`), although we'll want to include more reps when we evaluate our models. How many reps should we run?

lekoenig:
To get a better sense for how sensitive our model results are to the number of model reps, I re-trained the baseline LSTM model (version w/ 26 static features + 4 dynamic features) 25 times, bootstrapped samples of _n_ model reps, and calculated the variance of RMSE from each of those. 

Here's how I approached this question (using the `0_baseline_LSTM/exp_overall_metrics.csv` file and running the code below on both the `"train"` and `"val"` partitions):

```r
library(tidyverse)
set.seed(123)

# Define a function to bootstrap samples of the RMSE metric  
# (e.g. `n_samples` = 5, 10, 15 reps), for a given variable
# (e.g. 'do_mean') and partition (e.g. 'val'), resampling with
# replacement `n_resamples` times.
bootstrap_rmse <- function(overall_metrics, variable_select, partition_select, n_resamples, n_samples){
  
  metrics_subset <- overall_metrics |>
    filter(variable == variable_select, partition == partition_select)
  
  boot_samples <- bind_rows(lapply(seq_along(1:n_resamples), function(x){
    boot_samp <- sample(metrics_subset$rmse, n_samples, replace = TRUE)
    boot_samp_out <- tibble(variable = variable_select,
                            partition = partition_select,
                            resample_id = x,
                            n_samples = n_samples,
                            rmse_mean = mean(boot_samp))
    return(boot_samp_out)
  }))
  
  return(boot_samples)
}


boot_do_mean_train <- bind_rows(lapply(seq_along(1:25), function(x){
  bootstrap_rmse(overall_metrics, "do_mean", "train", n_resamples = 1000, n_samples = x)
})) 
boot_do_min_train <- bind_rows(lapply(seq_along(1:25), function(x){
  bootstrap_rmse(overall_metrics, "do_min", "train", n_resamples = 1000, n_samples = x)
})) 
boot_do_max_train <- bind_rows(lapply(seq_along(1:25), function(x){
  bootstrap_rmse(overall_metrics, "do_max", "train", n_resamples = 1000, n_samples = x)
}))
```

Below are two plots of the bootstrapped samples from the **train** partition. 

[...]

PR #165, opened 2022-10-14
Add static river/catchment attribute data

lekoenig:
This PR adds new river/catchment attributes to the file `1_fetch/in/target_sciencebase_attributes.csv` so that these features get downloaded, processed, and added to a formatted table that contains the static input feature datasets we're interested in (`p2_seg_attr_data`). 

The changes in this PR include the following steps:

- changes to `1_fetch/in/target_sciencebase_attributes.csv` to specify new attributes to collect. I gathered these based on my own hypotheses for features that might be important, but I expect we'll ultimately end up using only some portion of them. Please also feel free to suggest any other features that you think should be included ⭐ 
- processing the downloaded attributes. Most of the datasets are minimally processed since they are already referenced to the NHDPlusv2 network. However, we do further process the land cover data by aggregating the [NLCD index groups](https://www.mrlc.gov/data/legends/national-land-cover-database-class-legend-and-description) (e.g. 90 and 91) into reclassified groups (e.g. "wetland") based on the lookup table in `1_fetch/in/nlcd_landcover_reclassification.csv`. This file is new and is committed here so that others will have it when cloning the repo. 
- Combining all of the individual attribute data frames into a single table in `p2_seg_attr_data` using the function `combine_nhdv2_attr()`. 
- Summarizing the attribute values. As a quick method of QA/QC the pipeline now includes a target in `3_visualize` that outputs a csv file containing summary statistics (min/mean/max) for all of the processed attributes. I created a new function, `summarize_static_attributes()` to create and save that table.

This PR is just focused on downloading and formatting the attribute data. We'll use some method of feature selection to finalize the input variables to the LSTM models in #164.

 

Closes #51

PR #177, opened 2022-11-22
Input feature selection for static attributes

### LINKED ISSUE #164: Finalize input feature selection
Assess the variation in input variables across sites and use this information combined with mechanistic hypotheses/understanding to decide which input variables get added to the LSTM (more [here](https://github.com/USGS-R/drb-do-ml/discussions/170)).

lekoenig:
This PR adds new input features to the vector `x_vars_global` in the `_targets.R` pipeline config file, which specifies the input features that should be used across model runs. So far we've been using a "slim" set of input features that consists of 4 dynamic features representing meteorological driver inputs and 6 static features representing river/catchment attributes. The changes here include 30 total input features for the model to use when making predictions - 4 dynamic features and 26 static features. 

These features were selected from a set of variables added to the data pipeline in #165. The expanded set of features included as many as 108 static features, which I filtered down _manually_ based on correlations between features. For pairs of features that were highly correlated, I selected which one to retain based on my best guess of the feature most likely to mechanistically impact river dissolved oxygen concentrations. A more detailed description of the process I've taken is in #170. 

For this review, I'm looking for general feedback on the approach described in #170. I considered implementing a backwards selection procedure but am now leaning against it for a couple of reasons. First is that we're not necessarily striving to find the model with the *best* predictive accuracy. Second, given the feature importance plots I've shown, I think it's possible we'd be left with temperature and just a few other features, which might make it tough to generalize to new sites, including those in our test set. I think the decision to not include backward selection or another feature selection routine is OK since I've reduced correlated features in the set of 30 added here, although more features is perhaps less interpretable than a model with fewer input features.

I anticipate merging this PR and then pushing the trained weights in a separate PR as part of #176.

Closes #164

### DIFF
--- a/_targets.R
+++ b/_targets.R
@@ -96,8 +96,13 @@ test_start_date <- '2015-10-01'
 test_end_date <- '2022-10-01'
 
 # Define global model parameters for the "baseline" deep learning model
-x_vars_global <- c("pr","SLOPE","tmmx","tmmn","srad","CAT_BASIN_SLOPE","CAT_ELEV_MEAN",
-                   "CAT_IMPV11","CAT_CNPY11_BUFF100","CAT_TWI")
+x_vars_global <- c("tmmn","tmmx","pr","srad","SLOPE","TOTDASQKM","CAT_BASIN_SLOPE",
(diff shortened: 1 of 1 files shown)


jsadler2 (approved):
Looks great, @lekoenig.

Issue #158, opened 2022-09-08
Update container version


lekoenig:
There are currently two versions of the container on caldera: `do-ml_v0.2.sif` and `do-ml_v0.3.sif`. Perhaps if we're going to update the version to `v0.4` we should go ahead and update the container to add a few additional libraries that are used for visualization purposes, including {USAboundaries}, {ggspatial}, and {cowplot}.

lekoenig:
`{yaml}` seems to be installed on tallgrass, but the package version isn't recent enough to include some formatting options that I was expecting. I suggest we install this version if not the latest release:

```r
> packageVersion("yaml")
[1] ‘2.3.5’
```

jsadler2:
I've updated the docker image with those additional packages: jsadler2/do-ml:v0.5. So you should be able to now do `singularity pull docker://jsadler2/do-ml:v0.5` to get the latest.

lekoenig:
Thanks, Jeff! I pulled down the updated image and updated `launch-rstudio-container.slurm` to use the `do-ml_v0.5.sif` file. Is there a way to confirm that the new image is being used?

I tried to run `tar_make(p3_site_map_png)` but am still getting an error about the missing `USAboundaries` library.

PR #178, opened 2022-11-30
push trained model weights for 0_baseline_LSTM and 2_multitask_dense …

galengorski:
…with new features from #177. The only manual edit that I made was to change the `num_replicates` parameter from 1 to 10 in the `_targets.R` file, then I ran the model to create the weights


2a_model/src/models/0_baseline_LSTM/config.yml
```diff
@@ -1,8 +1,8 @@
-exp_name: "0_baseline_LSTM"
+exp_name: 0_baseline_LSTM
```
lekoenig, line 1: I specify that these should be quoted in the function that builds the config files so I was surprised to see that the quoted formatting gets lost when built on Tallgrass. I've played with this a bit before and I think if we updated the version of the `{yaml}` R package and added it to the container these config files would keep the quoted format as before. I've added a note in https://github.com/USGS-R/drb-do-ml/issues/158. 

It's good to know that the models still run even though these new config files don't have quotes around `exp_name` and `y_vars`.

lekoenig (approved):
This looks great - thank you, Galen! 

In https://github.com/USGS-R/drb-do-ml/pull/178/commits/0dbac60e45e7e63bc90c2d422a7c0bcff8d93ef9, I wondered whether we may as well keep the multitask models commented out in 2a_model.R since the training weights only include the 0_baseline_LSTM and the 2_metab_dense models. But I think it's fine to leave them - if I understand the snakemake workflow, when building the pipeline the models would not re-train for the baseline and metab_dense models, but would trigger training for the two multitask models. Is that right, @jsadler2?

lekoenig:
I've linked #176 to this PR since we've committed training weights for 10 model replicates here.

2a_model/src/models/0_baseline_LSTM/config.yml
jsadler2: I think python will automatically read these as strings since they aren't numbers

2a_model.R
```diff
@@ -189,14 +189,14 @@ p2a_targets_list <- list(
...
-             snakefile_dir = "0_baseline_LSTM",
-             config_path = stringr::str_remove(p2a_config_1a_metab_multitask_yml, "2a_model/src/models/")),
+         #the 1_ models use the same model and therefore the same Snakefile
+         #as the 0_baseline_LSTM run
+        #list(model_id = "1_metab_multitask",
+             #snakefile_dir = "0_baseline_LSTM",
+             #config_path = stringr::str_remove(p2a_config_metab_multitask_yml, "2a_model/src/models/")),
+        #list(model_id = "1a_multitask_do_gpp_er",
+             #snakefile_dir = "0_baseline_LSTM",
+             #config_path = stringr::str_remove(p2a_config_1a_metab_multitask_yml, "2a_model/src/models/")),
```
jsadler2, line 199: I commented these lines out since we didn't train new model weights for these models. 

This undoes 0dbac60e45e7e63bc90c2d422a7c0bcff8d93ef9. @galengorski  - was there a reason you uncommented the 1 and 1a models?

jsadler2 (approved):
I just have one question. Otherwise it looks good.

PR #181, opened 2022-12-16
plot model performance

lekoenig:
This PR adds a file containing a few plotting functions (`3_visualize/src/plotting_helpers.R`). I have not tried to incorporate these into the targets pipeline yet, and so for now to run these and generate the plots you can see `3_visualize/src/plotting_script_delete_later` where I lay out some pseudo-code that can ultimately be translated into targets.

Two plots are included for now:
baseline_rmse_plot

pgdl_schuyllkill_2


jsadler2 (approved):
These changes look good to me, @lekoenig.

PR #182, opened 2022-12-21
integrate functional performance

### LINKED ISSUE #179: Integrate Functional Performance into pipeline
To have the functional performace of all of the predictions, we will need to integrate the code into the pipeline.

jsadler2:
This PR integrates the functional performance calculations into the snakemake part of the pipeline. 

I mainly modified the function in `do_it_functions.py` file so it is calculating the FP for just one site, one sink, one source, and one replicate. That way, we can use Snakemake to run all of those combinations in parallel. 

closes #179


2a_model/src/do_it_functions.py
```diff
@@ -0,0 +1,183 @@
...
+    #because DO doesn't affect solar radiation, but it is needed for function
+    Mswap = np.stack((y_prepped, x_ss), axis = 1)
+    #x_bounds and y_bounds are for removing outliers
+    x_bounds = it_functions.find_bounds(M[:,0], 0.1, 99.9)
+    y_bounds = it_functions.find_bounds(M[:,1], 0.1, 99.9)
+    M_x_bound = np.delete(M, np.where((M[:,0] < x_bounds[0]*1.1) | (M[:,0] > x_bounds[1]*1.1)), axis = 0)
+    M_xy_bound = np.delete(M_x_bound, np.where((M_x_bound[:,1] < y_bounds[0]*1.1) | (M_x_bound[:,1] > y_bounds[1]*1.1)), axis = 0)
+
+    #calc it metrics and store in the dictionary it_dict
+    it_dict = it_functions.calc_it_metrics(M_xy_bound, Mswap, n_lags, nbins, calc_swap = False, alpha = 0.05, ncores = 7)
```
galengorski, line 133: I wonder if it might be better to break this function up and return `it_dict` instead of `max_it`. That way you could look at the TE, MI and correlation across different time lags before the rest of the function choose the `max_it` time lag from the dictionary.

galengorski (commented):
This all looks good @jsadler2 , I added one comment about outputting the intermediate `it_dict` dictionary object so that we have the correlations and IT metrics across different time lags to look at. We don't necessarily have to do that here, but I wanted to flag it and see what you thought.

2a_model/src/do_it_functions.py
jsadler2: Hm. I guess I don't understand what is in `it_dict` that we don't have in `max_it`. In line 162-163 we are storing the TE for all time steps 0-9 in `max_it` and in lines 168-169 we do the same for MI. Is that what you are referring to? 

If so, maybe `max_it` is actually just a bad variable name b/c it contains a lot more than the max TE and MI. 

If not, I think you are right in suggesting that we save more of the raw information somehow.

galengorski: Ok, so I think that `it_dict` has TE, MI, and correlation for all time lags (0-9) and `max_it` is a subset of `it_dict` where the maximum MI and TE values are chooses. That's how I wrote it and if I am reading the code right, I think that's what it's doing

galengorski: Ah ok, so I just got it running on my machine and I was mistaken, the `max_it` object does include the values of MI and TE for all time lags, it just identifies the maximum value for both modeled and observed and adds in the RMSE and makes the output dictionary a little easier to work with. So I think it's fine as is then, maybe adding a comment to explain what the `max_it` object contains might be helpful

jsadler2: Sounds good. I add those here: https://github.com/USGS-R/drb-do-ml/pull/182/commits/24c4b9cfdaf59cf5f4062f1eba212a121630756c

galengorski (approved):
Looks good to me!

PR #191, opened 2022-12-29
commit combined metric files

### LINKED ISSUE #189: commit combined metric csv files
I think it would be helpful and not to much of a size issue if we committed the combined results files to the repo. At least one or two of our main figures will come from these files and I think committing them would be helpful for at least two reasons. 

* First, we won't have to worry about rebuilding them. This could avoid some headaches that we've had before with getting nodes on HPC and/or unneeded/unwanted rebuilds. 
* Second, it would be nice to have a archived record of our metrics. We have past weights in the repo, but not metric files which makes it a little harder to reconstruct where we've been in the past if we need to do that. This will give us a checkpoint that we can look back on if we end up making changes to the model again (though I'm thinking we are pretty much done making changes to the model).

jsadler2:
committing combined metric files.

see #189 for more details/reasoning

closes #189


lekoenig (approved):
_From #189:_

> We have past weights in the repo, but not metric files which makes it a little harder to reconstruct where we've been in the past if we need to do that. This will give us a checkpoint that we can look back on if we end up making changes to the model again

I like the idea of committing the metric files mostly for this reason - once checked into version control, the metric files can now also serve as "log" files that enable us to see data changes more readily by inspecting the diffs. 

I see that the metrics files are created in `p2a_metrics_files.` The way I'm interpreting this target (and the snakemake workflow) is that if I run `tar_make()`, the pipeline will still build these files using the trained weights. But if the values in my built version differ from the values in these committed files, the files will show up as changed in git. Is that right, @jsadler2? In other words, I don't think `targets` will look for these files to decide whether `p2a_metrics_files` is outdated and needs to be built.

jsadler2:
> I like the idea of committing the metric files mostly for this reason - once checked into version control, the metric files can now also serve as "log" files that enable us to see data changes more readily by inspecting the diffs.

I hadn't thought of it this way, but I think you are right. 

> I see that the metrics files are created in p2a_metrics_files. The way I'm interpreting this target (and the snakemake workflow) is that if I run tar_make(), the pipeline will still build these files using the trained weights. But if the values in my built version differ from the values in these committed files, the files will show up as changed in git. Is that right, @jsadler2? In other words, I don't think targets will look for these files to decide whether p2a_metrics_files is outdated and needs to be built.

The metric files in this PR are actually created in `p2a_{overall,reach,FP}_metrics_file` targets (see #186). These depend on `p2a_metrics_files`. So if you pulled the repo and ran `tar_make()`, 
- targets should see that the model metrics files (e.g., `p2a_model/out/models/0_baseline_LSTM/exp_{overall,reach}_metrics.csv`) would not be there, so it would build the `p2a_metrics_files` target
- the `p2a_metrics_files` target will trigger Snakemake to use the committed trained weights to make predictions and then evaluate the predictions to produce the `exp_{overall,reach}_metrics.csv` and `{{model_id}}_func_perf.csv` files.
- then, since the upstream dependencies of `p2a_metrics_files` were updated, targets will rebuild the `p2a_{overall,reach,FP}_metrics_file` target which will overwrite the existing, committed overall metrics files (the ones in this PR). Assuming that everything works, the values should be exactly the same, so even though the files are technically overwritten, they should not be changed (which is what I think you were getting at with the log files).

jsadler2:
Not sure if that answers your question, @lekoenig, so let me know if it didn't

lekoenig:
> The metric files in this PR are actually created in p2a_{overall,reach,FP}_metrics_file targets (see https://github.com/USGS-R/drb-do-ml/pull/186). These depend on p2a_metrics_files

Ah, OK. I was looking at `p2a_metrics_files` [here](https://github.com/USGS-R/drb-do-ml/blob/main/2a_model.R#L236-L239), but I see now that the combined files are created in the targets below. I agree with your pipeline interpretation in the comment above - thanks for clarifying that. 

> Assuming that everything works, the values should be exactly the same, so even though the files are technically overwritten, they should not be changed (which is what I think you were getting at with the log files).

I agree. As long as the training weights are in `2a_model/out/models` these values should not differ. And yep, re. the log files, I think having these files committed is a nice check/confirmation on that assumption. If the values do change, the files will show up as changed in .git

galengorski:
This seems reasonable to me, not sure I have much to add on the targets integration, but I like the idea of having metrics in the repo

PR #192, opened 2023-01-24
add TE at zero lag functionality

galengorski:
This includes a small bit of code to add functionality to calculate transfer entropy (TE) at a time lag of zero. Previously we had been calculating TE by calculating the relationship between source (X) and sink (Y) as : 
image

where the time series of X and Y lagged at some time step tau. However, if tau = 0, then this formulation will lead to a TE of 0 because you are marginalizing out the whole time series. So the change that I propose is to use the above formulation for all tau not equal to zero and for tau = 0 to marginalize the t-1 lagged time series, so:

image

### DIFF
--- a/2a_model/src/it_functions.py
+++ b/2a_model/src/it_functions.py
@@ -208,14 +208,23 @@ def lag_data(M, shift):
     
     length_M = M.shape[0]
     cols_M = M.shape[1]
-
-    #this is for => H(Xt-T, Yt, Yt-T)
-    newlength_M = length_M - shift
-    M_lagged = np.nan*np.ones([newlength_M, cols_M+1])
-    M_lagged[:,0] = M[:(length_M-shift),0]
-    M_lagged[:,1] = M[shift:(length_M)+1,1]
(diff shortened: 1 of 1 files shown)


jsadler2:
@galengorski thanks for doing this. I looked over the code and ran it without error. I don't think I can really say if this is the "right" way to do it or not, though, since I don't really have experience with it. 

I did a sample calculation with the new code and this is what I get for one of the sites (01481500 and tmmx->do_min)

Figure_1

I'm a little surprised that correlation is actually less on day 0 than day -1. But it looks like the TE calculation reflects that for the observed and the 0_baseline_LSTM

galengorski:
That's an interesting result @jsadler2, maybe @jds485 might be able to weigh in on this discussion. It's encouraging to note that correlation and TE are showing maxima (absolute value) at the same time lag. I wonder how this translates to other sites?

jds485:
Galen and I discussed this. We think that the 2 equations should be presented as having different transfer entropy interpretations / use a different symbol to annotate the equations, but it is fine to use Y_t=1 when the lag is 0.

PR #202, opened 2023-02-27
198 add val data to training

jsadler2:
I moved the old validation data into the training set and I moved the old "test" set into the validation set. ⚠️  This PR will constitute the final results for the paper since we would no longer have a test set.

closes #198

PR #203, opened 2023-02-27
Little random edits

jsadler2:
**To be merged after #202**

A few minor edits.

* adding `1` and `1a` to the `add_weights.sh` bash file
* changing `ncores` to 1 in the `do_it_functions.py` file - I did this because I was having troubles with `multiprocessing` and snakemake both trying to do parallel scheduling at the same time
* adding the slurm file I've been using


jsadler2:
closed this b/c I want to simplify this.

PR #206, opened 2023-03-02
Filter out provisional data

lekoenig:
This PR omits provisional data from the DO data that is used for model training. Two new targets are added to subset the data downloaded from NWIS (`p2_daily_data_filtered` and `p2_inst_data_filtered`). I also updated downstream viz targets to use the new filtered data targets.

@jsadler2 We can either merge this or not. In practice, by updating `latest_date` to `2021-10-01` and with our current group of sites, the subset data are not different from the downloaded data because there are no provisional records. So these changes really just serve to codify our decision to not use provisional records in the code base. ⚠️ All targets downstream of `p2_daily_combined` would rebuild, which includes most if not all data prep targets.


jsadler2 (approved):
This looks good to me. I think it will be good to have this codified. Thanks, @lekoenig!

PR #220, opened 2023-08-09
Edit site map

### LINKED ISSUE #209: Add labels to study area figure
It might be helpful to include some labels in the study area figure to provide spatial context, e.g. major cities like Philadelphia.

lekoenig:
This PR includes code changes to make the following edits to the site map: 

- filter to the new set of 10 sites
- label Philadelphia
- add a base map to indicate urban areas/areas with high development
- label the sites with a site identifier

The base map is the 2019 NLCD impervious cover layer. To me this seemed like a nice way to show relative levels of urban density and still keep the base layer relatively simple. Note that this requires two new dependencies that are not in our dockerfile, {FedData} and {raster}. I'm not sure if it's possible to add these R libraries without updating the versions of any of the other libraries. I'm a little wary of updating the other packages at this stage so if that's not possible, maybe we consider taking this map out of the pipeline.

Regarding the site labels, I re-read the manuscript draft and what stood out to me as the biggest differentiator between sites was whether a site was located along the Schuylkill River or Brandywine Creek (the temporal holdout sites). I therefore opted to include site labels, including "SR" for the Schuylkill and "BC" for Brandywine. The number refers to the river mile (actually, river km) from the outlet and so can provide some information about river size if the site id is used in any of the other figures. What do you think?

edited_map

Draft caption text: 

> Figure 2. Monitoring sites used to train and test deep learning models that predict daily dissolved oxygen concentration in the lower Delaware River Basin, USA. Several sites are distributed along two tributaries to the Delaware River, Brandywine Creek (BC, n = 4) and the Schuylkill River (SR, n = 2). Two sites are located along Cobbs Creek in Philadelphia, PA (CC) and two sites are located along relatively forested, headwater streams in the northwest portion of the basin (Baptism Creek, BAP; French Creek, FC). Numbers within the site labels indicate the river distance (km) between the site and the river outlet. Red shading shows urban imperviousness, where darker shading indicates a higher percentage of urban impervious surface cover within a 30-m land surface pixel (NLCD 2019). 

Closes #209

### DIFF
--- a/2a_model.R
+++ b/2a_model.R
@@ -45,7 +45,8 @@ p2a_targets_list <- list(
         sf::st_as_sf(., coords = c("lon","lat"), crs = unique(.$epsg))
       
       append_river_km(site_splits, p1_nhd_reaches_sf) %>%
-        mutate(river_basin_abbr = case_when(river_basin == "Schuylkill" ~ "SR",
+        mutate(river_dist_km = round(river_dist_km, 0),
+               river_basin_abbr = case_when(river_basin == "Schuylkill" ~ "SR",
                                             river_basin == "Brandywine" ~ "BC",
                                             river_basin == "Cobbs" ~ "CC",
                                             site_id %in% c("014721254") ~ "FC",
--- a/2a_model.R
+++ b/2a_model.R
@@ -1,3 +1,4 @@
+source("2a_model/src/format_site_splits.R")
 source("2a_model/src/model_ready_data_utils.R")
 source("2a_model/src/write_model_config_files.R")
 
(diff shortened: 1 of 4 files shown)


lekoenig:
Do you think we should label the East- and West- branches of Brandywine Creek, e.g. WBBC_53.2?

galengorski:
I really like it @lekoenig I think the basemap does a nice job conveying information while also keeping it simple. For example, the reader will be able to see why the two different spatial holdouts are so different from looking at this map. I like the naming convention with the river and the rive km also, my only suggestion would be to simplify the names even further and round to the nearest km so `BC_23.6` -> `BC_24`. Also, maybe I missed it, but what does `BAP` stand for? Is there a site that is referring to? Other than that I think it looks awesome!

lekoenig:
> my only suggestion would be to simplify the names even further and round to the nearest km so BC_23.6 -> BC_24. 

Thanks, @galengorski! I agree with this and will round these river km to the nearest whole number. 📏 

> Also, maybe I missed it, but what does BAP stand for?

BAP stands for "BAPTISM CREEK" (014721259). Since the abbreviated "BC" would conflict with my Brandywine naming scheme, I just used this. Can you think of another site id that would be easier to interpret? I've edited the PR overview to include a draft caption - hopefully this helps, too. 

Baptism Creek is a relatively small stream that is very close to French Creek, the other headwater stream. At the scale of the lower DRB, these two sites are nearly indistinguishable from each other on the map, even if I used hollow circles.

lekoenig:
Good idea! Here's where we're at:

image

> Figure 2. Monitoring sites used to train and test deep learning models that predict daily dissolved oxygen concentration in the lower Delaware River Basin, USA. Several sites are distributed along two tributaries to the Delaware River, Brandywine Creek (BC, n = 4) and the Schuylkill River (SR, n = 2). Two sites are located along Cobbs Creek in Philadelphia, PA (CC) and two sites are located in close proximity along relatively forested, headwater streams in the northwest portion of the basin (Baptism Creek, BAP; French Creek, FC). Numbers within the site labels indicate the river distance (km) between the site and the river outlet. Red shading shows urban imperviousness, where darker shading indicates a higher percentage of urban impervious surface cover within a 30-m land surface pixel (NLCD 2019).

jsadler2:
I think this looks great, @lekoenig! I like the context of the urbanization and I like the simplified naming structure. 

I have a question and a suggestion

- **Question:**  Should we add some kind of indicator to the `CC` and `BAP` sites to indicate that they were the "Spatially Dissimilar" sites? It would be nice to not make the reader have to go back and forth in the text and the map to see which one's were "dissimilar." At the same time, when we introduce this map, we haven't talked about those experiments, so I wonder if that would be confusing since we haven't "gotten there" in the story yet. I guess after writing that out, I'd lean toward leaving them unmarked in this map, but I'd be curious to hear your guys' thoughts.
- **Suggestion:** Can we somehow indicate that the red is "urban imperviousness" in the figure itself instead of just in the caption? like this? 

259848375-9e1c5e57-d592-4b4a-9fb6-949cc3cebe9c

jsadler2:
@lekoenig I just tried to run this and got this error:
Screenshot 2023-08-21 at 11 50 35 AM

I think you just need to add the `2a_model/src/format_site_splits.R` file as part of the PR.

lekoenig:
Oops, sorry @jsadler2 - I just committed `2a_model/src/format_site_splits.R` along with changes to add the legend for impervious cover. Here is the map as it appears when I run the code in this PR:

image

